PR #11, opened 2016-09-15
Updating requirements.txt to use latest varcode, also fixing off-by-1 display error in report.

julia326:


vaxrank/report.py
```diff
@@ -71,8 +71,8 @@ def ascii_report_from_ranked_vaccine_peptides(
         for j, vaccine_peptide in enumerate(vaccine_peptides):
             mutant_protein_fragment = vaccine_peptide.mutant_protein_fragment
             amino_acids = mutant_protein_fragment.amino_acids
-            mutation_start = mutant_protein_fragment.mutant_amino_acid_start_offset
```
ihodes: Probably better for @iskandr to check this out, but was the problem that the "underlined" AA was one too far to the right? In which case this moves the underline one *more* to the right? I'm pretty unfamiliar with the offsets used around all of this, though...

ihodes:
btw, in a PR addressing an issue, you can write "fixes #10" (#10 being the issue), and when this PR is merged in, #10 will be automatically closed with a link to this PR.

julia326:
fixes #10

ihodes:
Hah! Commented right after me. Not sure "fixes" work in the comments, though? May have to be edited into the original comment at the top?

vaxrank/report.py
julia326: The "underlined" AA was one too far to the left actually. So the new output becomes, from the test script:

1) chr9 g.82927102G>T (Phip)
	Top score: 3.69
	Predicted effect: Substitution(variant=chr9 g.82927102G>T, transcript_name=Phip-003, transcript_id=ENSMUST00000034787, effect_description=p.T441N)
	Reads supporting variant allele: 17
	Reads supporting reference allele: 39
	Reads supporting other alleles: 0
	Vaccine Peptides:
		i. DNTVIHAV_N_NMTLKVWNSYTGQL (score = 3.69)
		   - Length: 23
		   - Expression score: 4.12
		   - Mutant epitope score: 0.90
		   - Wildtype epitope score: 1.03
		   - Reads fully spanning cDNA sequence(s): 2
		   - Mutant amino acids: 1
		   - Mutation distance from edge: 7
		   - Predicted mutant epitopes:
			 * HAVNNMTLKV 202.58 (H-2-Db)
			 * VIHAVNNM 714.63 (H-2-Kb)
			 * HAVNNMTL 1077.76 (H-2-Db)
			 * IHAVNNMTLKV 1968.09 (H-2-Db)

julia326:
ah makes sense, first time using git pull requests :)

vaxrank/report.py
ihodes: Perfect; this looks good to me! Is there a comment you could stick there explaining why we need to add 1?

julia326: done! how do PR approvals work? verbal LGTM or some tool-y way of doing it?

vaxrank/report.py
```diff
@@ -71,8 +71,10 @@ def ascii_report_from_ranked_vaccine_peptides(
...
-            mutation_end = mutant_protein_fragment.mutant_amino_acid_end_offset
+            # Fixes an off-by-1 error for displaying the peptide; previously
+            # the highlighted mutation was one too far to the left.
+            mutation_start = mutant_protein_fragment.mutant_amino_acid_start_offset + 1
```
iskandr: This fixes the display bug but worryingly the offset is still off by 1. Do you mind if I add some unit tests to this branch to sanity check `mutant_protein_fragment`?

iskandr:
We usually just say "LGTM" in the thread.

vaxrank/report.py
julia326: ah my bad, didn't realize this was a more-than-display bug. I'm happy to add a unit test, though need some better understanding of how this works and what the components are - let's chat offline

iskandr:
Closing this in favor of #12

PR #12, opened 2016-09-16
Added tests to make sure mutant amino acids in MutantProteinFragment match varcode prediction

iskandr:
* Copied B16 mouse data from isovar
* Removed some print statements and converted others to `logging.info`
* Using Ensembl release 85 instead of 84
* Updated varcode dependency to newer version

Closes https://github.com/hammerlab/vaxrank/issues/10


ihodes:
LGTM!

PR #18, opened 2016-09-22
Adding vaxrank HTML/PDF output, with some associated changes

### LINKED ISSUE #3: Generate HTML and PDF reports
Currently only making a text file for vaccine peptides, would be nice to format this as HTML and then convert to PDF with [pdfkit](https://github.com/JazzCore/python-pdfkit)

julia326:
- refactored report generation to create a dictionary of values which then gets fed into ASCII/HTML/whatever template
- added an HTML template and an ASCII template that's very close to previous output (only difference should be whitespace)
- added report outputs to .gitignore
- updated some requirements files

Fixes #3 - also see PDF for what this looks like

[vaccine-peptides-report.pdf](https://github.com/hammerlab/vaxrank/files/488144/vaccine-peptides-report.pdf)

### DIFF
--- a/run-vaxrank-b16-test-data.sh
+++ b/run-vaxrank-b16-test-data.sh
@@ -4,4 +4,5 @@ vaxrank \
     --vaccine-peptide-length 25 \
     --mhc-predictor netmhc \
     --mhc-alleles H2-Kb,H2-Db \
-    --padding-around-mutation 5 --output-ascii-report vaccine-peptides-report.txt
+    --padding-around-mutation 5 \
+    --output-ascii-report vaccine-peptides-report.txt
--- a/run-vaxrank-b16-test-data.sh
+++ b/run-vaxrank-b16-test-data.sh
@@ -4,4 +4,4 @@ vaxrank \
     --vaccine-peptide-length 25 \
     --mhc-predictor netmhc \
     --mhc-alleles H2-Kb,H2-Db \
-    --padding-around-mutation 5
+    --padding-around-mutation 5 --output-ascii-report vaccine-peptides-report.txt
--- a/vaxrank/cli.py
+++ b/vaxrank/cli.py
@@ -25,12 +25,14 @@ from mhctools.cli import (
 from isovar.cli.variant_sequences import make_variant_sequences_arg_parser
 from isovar.cli.rna_reads import allele_reads_generator_from_args
 
-
 from .core_logic import ranked_vaccine_peptides, dataframe_from_ranked_list
 from .report import (
-    make_ascii_report_from_ranked_vaccine_peptides,
-    make_html_report_from_ranked_vaccine_peptides,
+    compute_template_data,
+    make_ascii_report,
+    make_html_report,
+    make_pdf_report,
 )
+
 
 # inherit all commandline options from Isovar
 arg_parser = make_variant_sequences_arg_parser(
@@ -49,17 +51,17 @@ arg_parser.add_argument(
 
 arg_parser.add_argument(
     "--output-ascii-report",
-    default="vaccine-peptides-report.txt",
+    default="",
     help="Path to ASCII vaccine peptide report")
 
 arg_parser.add_argument(
     "--output-html-report",
-    default="vaccine-peptides-report.html",
+    default="",
     help="Path to HTML vaccine peptide report")
 
 arg_parser.add_argument(
     "--output-pdf-report",
-    default="vaccine-peptides-report.pdf",
+    default="",
     help="Path to PDF vaccine peptide report")
 
 vaccine_peptide_group = arg_parser.add_argument_group("Vaccine peptide options")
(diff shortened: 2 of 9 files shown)


vaxrank/report.py
```diff
@@ -76,37 +86,101 @@ def ascii_report_from_ranked_vaccine_peptides(
...
+        bam_path,
+        ascii_report_path):
+    template = JINJA_ENVIRONMENT.get_template('templates/template.txt')
+    template_data = compute_template_data(
+        ranked_variants_with_vaccine_peptides,
+        mhc_alleles,
+        variants,
+        bam_path)
+    report = template.render(template_data)
+    print(report)
```
ihodes: I think we probably shouldn't print the report, though I know we were doing this before

vaxrank/cli.py
```diff
@@ -120,13 +133,17 @@ def main(args_list=None):
...
         variants=variants,
-        bam_path=args.bam)
+        bam_path=args.bam,
+        ascii_report_path=args.output_ascii_report)
 
-    print(ascii_report)
-
-    with open(args.output_ascii_report, "w") as f:
-        f.write(ascii_report)
+    make_html_report_from_ranked_vaccine_peptides(
```
ihodes: Might be nice if all report are optional, so we don't litter the directory with 3 report.html/txt/pdf? 

I'd imagine we could require 1+ of "--output-X" options.

vaxrank/report.py
```diff
@@ -76,37 +86,101 @@ def ascii_report_from_ranked_vaccine_peptides(
...
+    length = 8  # 8" baseline from patient info header
+    variants = template_data['variants']
+    for variant in variants:
+        length += 6  # 6" space for variant description
+        for peptide in variant['peptides']:
+            length += 4  # 4" space for peptide description
+    return length
+
+
+def make_html_report_from_ranked_vaccine_peptides(
```
ihodes: Could this be split into make_pdf_report… (which I guess would just call make_html_… making a tmp html report, and then convert it to PDF, then remove the html file); would be nice to have them all split up.

vaxrank/cli.py
hammer: 👍  for optional reports. Ultimately our reports will have information coming from sources other than vaxrank so we'll want some way to grab structured output from vaxrank and not the report itself.

vaxrank/report.py
julia326: yep gp - changing to a logging statement instead

run-vaxrank-b16-test-data.sh
```diff
@@ -4,4 +4,4 @@ vaxrank \
     --vaccine-peptide-length 25 \
     --mhc-predictor netmhc \
     --mhc-alleles H2-Kb,H2-Db \
-    --padding-around-mutation 5
+    --padding-around-mutation 5 --output-ascii-report vaccine-peptides-report.txt
```
ihodes: Definitely a nit, but maybe line-break this to keep it in line with the rest of the file?

vaxrank/cli.py
```diff
@@ -133,17 +135,23 @@ def main(args_list=None):
...
+    template_data = compute_template_data(
         ranked_variants_with_vaccine_peptides=ranked_list,
         mhc_alleles=mhc_alleles,
         variants=variants,
-        bam_path=args.bam,
-        html_report_path=args.output_html_report,
-        pdf_report_path=args.output_pdf_report)
+        bam_path=args.bam)
+
+    if args.output_ascii_report:
```
ihodes: If no report types are specified, we should fail fast and require one to be (otherwise the user will just wait a bit and Vaxrank will run, but nothing will happen!)

vaxrank/report.py
```diff
@@ -154,23 +163,14 @@ def compute_pdf_length_from_template_data(template_data):
...
-        ranked_variants_with_vaccine_peptides,
-        mhc_alleles,
-        variants,
-        bam_path)
-    with open(html_report_path, "w") as f:
-        f.write(template.render(template_data))
+def make_pdf_report(
+        template_data,
+        pdf_report_path):
+    path = "%s.html" % uuid.uuid4()
```
ihodes: You might like this context manager ✨  https://docs.python.org/3/library/tempfile.html#tempfile-examples

ihodes:
If you rebase this on master, does Travis still fail? It's related to the recent biotypes changes @iskandr just made.

run-vaxrank-b16-test-data.sh
julia326: hilariously enough, I did actually do this and try to push to master - but Gogo gave up on me literally 1 second before the 'git push' went through :) the lack of line break was unbearable when I realized it

vaxrank/report.py
julia326: I tried to make it work with tempfile first, but pdfkit didn't seem to like reading from a temp HTML file. I'll play around with it some more, probably a weird permissions thing

timodonnell:
Just a random thought in case it turns out to be related to the tempfile
issue: if you are using tempfile.NamedTemporaryFile and haven't closed the
file handle after writing to it, you'll want to call .flush() on the handle
before passing its name to another tool. Otherwise the output may not have
made it into the file yet due to buffering.

julia326:
@timodonnell that was indeed an issue, thank you!

ihodes:
LGTM! Merge away 👏  First PR = major need for the pipeline.

PR #33, opened 2016-09-27
{py}pandoc note in RELEASING.md; upping version to fix readme markdown

julia326:


arahuja:
LGTM! - looks like `conda install -c conda-forge pypandoc` installs both `pypandoc` and `pandoc` and works for Travis

PR #50, opened 2016-10-06
One logger per module

julia326:
- changing some verbose INFOs to DEBUGs
- adding a logging.conf file: one place to configure all logs


iskandr:
LGTM

PR #55, opened 2016-10-17
Add manufacturability scores to vaccine peptide ranking

iskandr:
Added manufacturability scores to vaccine peptide ranking based on discussions with JPT. Also filtering out MHC binding predictions near 2000nM with a `--min-epitope-score` option that defaults to `0.0001`.

Aside: I'm concerned that @julia326 and I are duplicating code that turns VaccinePeptide objects into lists of dictionaries. I wonder if the report and DataFrame->CSV code can be somehow combined. I'm also wondering if the repetition between the text and HTML reports can be eliminated.


vaxrank/vaccine_peptide.py
```diff
@@ -54,14 +60,35 @@ def lexicographic_sort_key(self):
...
+            # All else being equal, we prefer to maximize the number of
+            # mutant amino acids
             -self.mutant_protein_fragment.n_mutant_amino_acids,
+            # If nothing else can serve as a tie break then try to center
+            # the mutation in the vaccine peptide.
             -self.mutant_protein_fragment.mutation_distance_from_edge
         )
+        return (
+            essential_score_tuple +
+            manufacturability_score_tuple +
```
ihodes: Are we sure we want ease of manufacturability to outweigh potentially immunogenic concerns? I can see dropping peptides we can't hope to manufacture, but if we're not sure we can't, why sort them higher than peptides we might prefer to manufacture?

ihodes:
LGTM % 1 question!

vaxrank/vaccine_peptide.py
iskandr: It seems like there are at least two immunogenic concerns here: (1) minimizing the wildtype epitope score and (2) the unfortunate fact that [TCR contact residue hydrophobicity is a hallmark of immunogenic CD8+ T cell epitopes.](https://www.ncbi.nlm.nih.gov/pubmed/25831525). 

One thing I can imagine doing is splitting the hydrophobicity scores tuple into higher priority binary scores (C-term 7mer GRAVY > 1.5, max kmer GRAVY >1.5) and then a lower priority continuous value that we place below wildtype epitope score. 

What do you think?

JPFinnigan: Hey, 

I'll think about this and get back to you. I would lean towards warning the end-user about manufacturability concerns (e.g. "WARNING - Very Hydrophobic C-terminus"), and suggesting alternatives rather than having "manufacturability" factor into ranking. Peptide synthesis is an evolving science, and you probably don't not want to bake-in current synthesis limitations. For example, the reactive residues JPT suggests we avoid (e.g. Cysteine) are useful handles for creating peptide-conjugates, which many groups are exploring.

iskandr: Thanks for replying @JPFinnigan. We've talked about the wisdom of baking in manufacturability when we might want to eventually use other peptide synthesis methods or even mRNA vaccines. I think the right approach is to give the user options for *which* manufacturability scores to use since there will probably (at the very least) always be some undesirable N and C terminal residues. So, think of this as a first version which will get generalized once we want to use some other kind of peptide synthesis and/or vaccine.

iskandr:
@julia326 @armish @ihodes Want to take a second look?

iskandr:
Closes https://github.com/hammerlab/vaxrank/issues/2

iskandr:
I want this to be very explicit (a default parameter has a way of getting
silently changed).

Issue #48, opened 2016-10-04
test_mutant_protein_sequence.py is flaky


julia326:
Sometimes fails with:

```
vaxrank/vaxrank/core_logic.py:91: RuntimeWarning: invalid value encountered in double_scalars
  if vp.combined_score / max_score > 0.99
E..
======================================================================
ERROR: test.test_mutant_protein_sequence.test_mutant_amino_acids_in_mm10_chr9_82927102_refGT_altTG_pT441H
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/julia/Envs/vaxrank/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/julia/code/vaxrank/test/test_mutant_protein_sequence.py", line 91, in test_mutant_amino_acids_in_mm10_chr9_82927102_refGT_altTG_pT441H
    vaccine_peptide = vaccine_peptides[0]
IndexError: list index out of range
```

iskandr:
That's really confusing! Looks like it might be related to floating point warning?

iskandr:
@julia326 Just a reminder to myself that we need to understand this error (never observed on my machine, we should figure out when/how it happens).

julia326:
@iskandr agreed - just happened again on Travis:

https://travis-ci.org/hammerlab/vaxrank/jobs/168180329

let's look at it tomorrow?

PR #59, opened 2016-10-19
Vaxrank self peptide predictor

### LINKED ISSUE #28: Filter self peptides
Bring ligandome back to life!

julia326:
Fixes #28, #16, #48

### DIFF
--- a/test/test_epitope_prediction.py
+++ b/test/test_epitope_prediction.py
@@ -1,0 +1,59 @@
+# Copyright (c) 2016. Mount Sinai School of Medicine
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import, print_function, division
+
+from argparse import ArgumentParser
+from nose.tools import eq_, ok_
+
+from mhctools import RandomBindingPredictor
+from pyensembl import EnsemblRelease
+from vaxrank.epitope_prediction import predict_epitopes
+from vaxrank.mutant_protein_fragment import MutantProteinFragment
+from vaxrank.vaccine_peptide import VaccinePeptide
+
+
+def test_reference_peptide_logic():
+    protein_fragment = MutantProteinFragment(
+    	variant=None,  # irrelevant to test
+    	gene_name=u'Wdr13',
+    	amino_acids='KLQGHSAPVLDVIVNCDESLLASSD',
+    	mutant_amino_acid_start_offset=12,
+    	mutant_amino_acid_end_offset=13,
+    	n_overlapping_reads=71,
+    	n_alt_reads=25,
+    	n_ref_reads=46,
+    	n_alt_reads_supporting_protein_sequence=2)
+
+    epitope_predictions = predict_epitopes(
+        mhc_predictor=RandomBindingPredictor(["H-2-Kb", "H-2-Db"]),
+        protein_fragment=protein_fragment,
+        genome=EnsemblRelease(species="mouse"))
+
+    # occurs in protein ENSMUSP00000033506
+    prediction_occurs_in_reference = epitope_predictions['NCDESLLAS']
+    prediction_does_not_occur_in_reference = epitope_predictions['LDVIVNCDE']
+    ok_(prediction_occurs_in_reference.occurs_in_reference)
+    ok_(not prediction_does_not_occur_in_reference.occurs_in_reference)
+
+    # construct a simple vaccine peptide having these two predictions, which makes it easy to check
+    # for mutant/WT scores from single contributors
+    vaccine_peptide = VaccinePeptide(
+    	protein_fragment,
+    	[prediction_occurs_in_reference, prediction_does_not_occur_in_reference])
+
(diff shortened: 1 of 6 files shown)


vaxrank/epitope_prediction.py
```diff
@@ -77,24 +84,52 @@ def logistic_score(
...
+    logger.info("Computing self peptides...")
+    logger.info(genome)
+    # check if this exists
+    kmer_dict = defaultdict(set)
+    for k in ks:
+        for t in genome.transcripts():
+            if t.is_protein_coding:
+                p = t.protein_sequence
+                if p:
+                    for i in range(len(p) - k):
```
iskandr: If `p` is 9 amino acids and we want to generate reference 9mers then this expression would evaluate to `range(9 - 9)` = `range(0)`. Should this be `len(p) - k + 1`? I think this is my mistake from when we were prototyping earlier.

iskandr:
Looks good so far. What are you think for disk caching of the generated peptide sets?

vaxrank/epitope_prediction.py
```diff
@@ -77,24 +84,52 @@ def logistic_score(
...
         return logistic / normalizer
 
-def predict_epitopes(mhc_predictor, protein_fragment):
+# TODO(julia): move this code to probably topiary
+def compute_self_peptides(genome, ks):
+    logger.info("Computing self peptides...")
+    logger.info(genome)
+    # check if this exists
+    kmer_dict = defaultdict(set)
+    for k in ks:
```
julia326: swapped, though not sure where the "twice" comes from

vaxrank/epitope_prediction.py
julia326: yep good call, fixing

vaxrank/epitope_prediction.py
```diff
@@ -77,24 +86,86 @@ def logistic_score(
...
+        with open(cache_path(genome, k)) as f:
+            kmer_dict[k] = set(f.read().split('\n'))
+    logger.info("Done.")
+    return kmer_dict
+
+def predict_epitopes(mhc_predictor, protein_fragment, genome=None):
     results = []
+    self_peptides = load_self_peptides(genome, (8, 9, 10))
```
iskandr: Instead of hard-coded epitope lengths we should use whatever the user picks via a commandline arg.

vaxrank/epitope_prediction.py
```diff
@@ -77,24 +87,84 @@ def logistic_score(
...
+    -------
+    fm : shellinford.FMIndex
+        Index populated with reference peptides from the genome
+    """
+    path = fm_index_path(genome)
+    if force_reload or not os.path.exists(path):
+        logger.info("Loading FM index to %s", path)
+        fm = shellinford.FMIndex()
+        fm.build(
+            [t.protein_sequence for t in genome.transcripts() if t.is_protein_coding],
```
ihodes: I'd make this a generator (with `()` instead of a list with `[]`) to save on compute/memory resources.

test/test_epitope_prediction.py
```diff
@@ -0,0 +1,67 @@
...
+    	n_ref_reads=46,
+    	n_alt_reads_supporting_protein_sequence=2)
+
+    epitope_predictions = predict_epitopes(
+        mhc_predictor=RandomBindingPredictor(["H-2-Kb", "H-2-Db"]),
+        protein_fragment=protein_fragment,
+        genome=EnsemblRelease(species="mouse"))
+
+    # spot-check a couple of peptides on predictions
+    prediction_occurs_in_reference = find_prediction_by_sequence(
```
iskandr: Would it be easier if `predict_epitopes` returned an `OrderedDict`?

vaxrank/core_logic.py
```diff
@@ -29,6 +29,11 @@
 logger = logging.getLogger(__name__)
 
 
+# Lifted from https://www.python.org/dev/peps/pep-0485/#proposed-implementation
+def _isclose(a, b, rel_tol=1e-09, abs_tol=0.0):
```
iskandr: Probably better to reuse https://docs.scipy.org/doc/numpy/reference/generated/numpy.isclose.html rather than implement this via copy/paste.

vaxrank/core_logic.py
```diff
@@ -83,15 +95,18 @@ def vaccine_peptides_for_variant(
...
+        return candidate_vaccine_peptides
 
+    max_score = max(vp.combined_score for vp in candidate_vaccine_peptides)
     # only keep candidate vaccines that are within 1% of the maximum
     # combined score
     filtered_candidate_vaccine_peptides = [
         vp
         for vp in candidate_vaccine_peptides
-        if vp.combined_score / max_score > 0.99
+        if _isclose(max_score, 0.0) or vp.combined_score / max_score > 0.99
```
iskandr: Maybe do this check outside the list comprehension?

```
if np.isclose(max_score, 0.0):
    filtered_vaccine_peptides = candidate_vaccine_peptides
else:
    filtered_vaccine_peptides = ... # old list comprehension
```

I also don't remember at which point we would filter out vaccine peptides whose score is 0.

vaxrank/epitope_prediction.py
```diff
@@ -77,34 +87,101 @@ def logistic_score(
...
+        Index populated with reference peptides from the genome
+    """
+    path = fm_index_path(genome)
+    if force_reload or not os.path.exists(path):
+        logger.info("Loading FM index to %s", path)
+        fm = shellinford.FMIndex()
+        fm.build(
+            (t.protein_sequence for t in genome.transcripts() if t.is_protein_coding),
+            path)
+        logger.info("Done.")
```
iskandr: When you're logging "Done" any chance you could include what got done (e.g. "Done building FM index")? Sometimes I see those in the logs and it's not obvious to me.

julia326: can do that, though it reloads very quickly so probably doesn't save that much time

ihodes:
LGTM!

PR #65, opened 2016-10-25
Refactor report ordered dict

### LINKED ISSUE #64: Refactor report gen to use OrderedDicts
to avoid replication in all the templates

julia326:
A couple of report gen changes:
    - Refactored report gen logic to use OrderedDicts. This brings all the work into the Python code, and avoids replication across templates, which now mostly just iterate over the OrderedDicts
    - Prettied up the peptide table; now "Manufacturability" and "Predicted mutant epitopes" are both subsections.

Fixes #64

### DIFF
--- a/vaxrank/report.py
+++ b/vaxrank/report.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from __future__ import absolute_import, print_function, division
+from __future__ import absolute_import, division
 from collections import OrderedDict
 import logging
 import os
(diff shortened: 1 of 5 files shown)


vaxrank/report.py
```diff
@@ -63,85 +67,110 @@ def compute_template_data(
...
-                if epitope_prediction.overlaps_mutation:
-                    score = epitope_prediction.logistic_score()
-                    epitope_dict = {
-                        'sequence': epitope_prediction.peptide_sequence,
-                        'ic50': epitope_prediction.ic50,
-                        'normalized_binding_score': score,
-                        'allele': epitope_prediction.allele,
-                    }
-                    epitopes.append(epitope_dict)
+        ############################### Variant info ###############################
```
ihodes: What about factoring out all these Foo Infos into functions like `foo_info(...)`. Would be a bit easier to read and follow that way, I think?

ihodes:
LGTM!

PR #70, opened 2016-10-28
Vaxrank report changes:

### LINKED ISSUE #58: Add CSV report(s)
CSV per variants w/ N vaccine peptides in it

julia326:
- added manufacturer report gen and consolidated with existing --output-csv functionality
- what's getting saved now is the ranked variant list output of core_logic.py, rather than processed template data
- a bit of manufacturability tuple creation refactoring to make it clearer (from Alex)
- refactored cli.py substantially to split up the various components, also make it easier/cleaner to add new non-template report outputs
- couple logging changes:
  - added a file log to default logging config (helps with grepping/debugging locally)
  - logging line numbers

This is a bit ugly, because there are various old/new parsed args going around, but I couldn't figure out a way to change it without changing how we create MHC alleles/variant collections/getting rid of pickling altogether/etc.

Fixes #58, #68


vaxrank/cli.py
```diff
@@ -41,144 +44,148 @@
...
+            "expression data, and patient HLA type."),
+    )
+    add_mhc_args(arg_parser)
+    add_vaccine_peptide_args(arg_parser)
+    add_output_args(arg_parser)
+    return arg_parser
+
+
+def cached_run_arg_parser():
+    arg_parser = ArgumentParser()
```
iskandr: Shouldn't this get the same prog and description as above?

vaxrank/cli.py
```diff
@@ -13,26 +13,29 @@
 # limitations under the License.
 
 from __future__ import absolute_import, print_function, division
+from collections import OrderedDict
 import sys
 import logging
 import logging.config
 import pickle
```
iskandr: Consider using six.moves.cPickle

vaxrank/cli.py
```diff
@@ -190,38 +197,108 @@ def load_template_data(args):
...
+        'saved_args': args,
+    }
+    logger.info('About to save args: %s', data['saved_args'])
+
+    # save pickle data if necessary. as of time of writing, vaxrank takes ~25 min to run,
+    # most of which is core logic. the formatting is super fast, and it can
+    # be useful to save the data to be able to iterate just on the formatting
+    if args.output_pickle_file:
+        with open(args.output_pickle_file, 'wb') as f:
+            pickle.dump(data, f)
```
iskandr: Make the protocol explicit?

julia326: Any explicit value other than 0 fails. Here is what happens at deserialization, when I dump with protocol=2:

```
  File "/Users/julia/code/vaxrank/vaxrank/cli.py", line 186, in ranked_variant_list_with_saved_args
    data = cPickle.load(f)
TypeError: __new__() takes at most 4 arguments (6 given)
```
Would like to leave it unspecified for now - sadly, can't test on py3 right now because FMIndex is busted on py3 so it errors for other reasons.

vaxrank/cli.py
julia326: done

vaxrank/cli.py
julia326: done

vaxrank/cli.py
```diff
@@ -239,32 +316,20 @@ def main(args_list=None):
...
-            "--output-html-report, "
-            "--output-pdf-report")
-
-    template_data = load_template_data(args)
-
-    if args.output_ascii_report:
-        make_ascii_report(
-            template_data=template_data,
-            ascii_report_path=args.output_ascii_report)
+    if "--input-pickle-file" in args_list:
```
julia326: Not sure what you mean. Would you want to see a separate binary? (That's what I was trying to figure out earlier, ie how you'd want to interact with the vaxrank CLI in the cases of a pickled load)

vaxrank/manufacturability.py
```diff
@@ -122,25 +122,25 @@ def cysteine_count(amino_acids):
...
-    distant parts of the peptide,
+    distant parts of the peptide
     """
     return sum(amino_acid == "C" for amino_acid in amino_acids)
 
 def combine_scoring_functions(*scoring_functions):
     """
     Given a list of scoring functions, make a namedtuple with
-    fields of the same names.
+    fields of the same names. Returns a ManufacturabilityScores object.
```
julia326: done

vaxrank/cli.py
julia326: Now that FMIndex on py3 is fixed (amazing turnaround time on that shellinford issue!), pickling in py3 fails at serialization with:

```
Traceback (most recent call last):
  File "/Users/julia/Envs/vaxrank3/bin/vaxrank", line 11, in <module>
    load_entry_point('vaxrank', 'console_scripts', 'vaxrank')()
  File "/Users/julia/code/vaxrank/vaxrank/cli.py", line 333, in main
    data = ranked_variant_list_with_saved_args(args)
  File "/Users/julia/code/vaxrank/vaxrank/cli.py", line 216, in ranked_variant_list_with_saved_args
    cPickle.dump(data, f)
AttributeError: Can't pickle local object 'combine_scoring_functions.<locals>.ManufacturabilityScores'
```

Unsurprisingly, this fails independent of protocol. So the issue is that pickling doesn't play nice with `ManufacturabilityScores`, and `serializable` doesn't play nice with `argparse` objects. At this point, there are a couple of options:
- make ManufacturabilityScores an actual namedtuple or class, and not construct it on the fly (stick with pickling)
- don't save args from a previous run, when reloading
- look into rewrite `mhc_alleles_from_args`, `variant_collection_from_args` to be constructed from dicts instead of `argparse.Namespace` objects - this probably wouldn't be so bad - and switch from pickling to `serializable` (kinda overloading `serializable` some more)

@alex thoughts?

vaxrank/report.py
```diff
@@ -197,11 +216,14 @@ def compute_template_data(self):
...
             variant_short_description = variant.short_description
             if len(vaccine_peptides) == 0:
-                logger.info("Skipping %s, no vaccine peptides", variant_short_description)
+                logger.info("Skipping gene(s) %s, variant %s: no vaccine peptides",
+                    variant.gene_names, variant_short_description)
                 continue
 
             variant_data = self._variant_data(vaccine_peptides[0])
-            predicted_effect = self.variants_to_top_coding_effect_dict.get(variant)
+            # TODO(julia): is this right?
```
julia326: @iskandr you recently added `supporting_reference_transcripts` as a property on the `MutantProteinFragment`, which is connected to a particular peptide. but when we display a predicted effect, it's variant-level as opposed to peptide-level. what's the best thing to do here? do you take the top peptide and grab its supporting transcripts? or do what i'm doing here now?

PR #72, opened 2016-10-31
cache pyensembl directory during Travis tests

iskandr:


.travis.yml
```diff
@@ -3,6 +3,12 @@ language: python
 python:
   - "2.7"
   - "3.4"
+cache:
+  pip: true
+  # cache directory used for Ensembl downloads of GTF and FASTA files
+  # along with the indexed db of intervals and ID mappings and pickles
+  # of sequence dictionaries
+  directories: /home/travis/.cache/pyensembl/
```
julia326: is it also worth caching the FM indices? that'll live at `/home/travis/.cache/vaxrank`, though I can rename the directory to something more descriptive

iskandr: Probably? I'm less clear on how often the FM index format might change or what will happen if the old file is invalid. 

Very incompletely considered thought: cache the FM index as long as you also delete in in vaxrank if/when the file doesn't open correctly.

PR #81, opened 2016-11-09
Fixing issue where we were dropping peptides that bound to multiple alleles

### LINKED ISSUE #80: Dropping peptides that bind to multiple alleles from epitope predictions
Simple bug using a dict poorly, fixing now.

julia326:
Verified that this fixes the report

Fixes #80


ihodes:
LGTM!

PR #106, opened 2017-01-12
shortening sheet name in Excel output

iskandr:
Fixes https://github.com/hammerlab/vaxrank/issues/103


ihodes:
LGTM! Thanks for the fix.

Issue #114, opened 2017-02-16
Vaxrank fails at saving results when there are no vaccines available


armish:
Since `4.0.0`, a majority of my vaxrank runs within epidisco fail with the following error and I am assuming that is because for these failed run, there is no potential vaccine available:

```
[...]

Issue #24, opened 2016-09-25
Add unit tests for generating HTML/ASCII/PDF reports


iskandr:

iskandr:
I added HTML and CSV output generation (accidentally pushed to master). We should add ASCII (and maybe PDF) and then check that all reports contain the same vaccine peptides.

iskandr:
@julia326 We're now testing the report code, right?

ihodes:
@julia326 can we close this issue now?

PR #116, opened 2017-03-03
Fix crashing report when there are no variants

### LINKED ISSUE #115: Omit peptides with no mutant epitopes
Right now such a vaccine peptide will still show up with a blank section for "predicted mutant epitopes".

julia326:
A few changes:
- filtering out variants with no peptides, and peptides with no mutant epitopes
- fixing crash in case where there are no variants, with a test
- adding basic test cases for all report formats

Fixes #115 , #114, #24


armish:
thank you for the fix, @julia326! LGTM :shipit:

PR #117, opened 2017-03-06
Fix mhctools usage

iskandr:
* Synchronized code that uses mhctools `BindingPrediction` objects with changes to that library.
* Simplified `EpitopePrediction` by getting rid of `EpitopePredictionBase`
* `logistic_score` method -> `logistic_epitope_score` helper function
* Fixed intermittent failure of shellinford on Py2 by converting strings from `unicode` to `str`

Closes https://github.com/hammerlab/vaxrank/issues/78


vaxrank/epitope_prediction.py
```diff
@@ -25,68 +25,47 @@
...
+    "ic50",
+    "percentile_rank",
+    "prediction_method_name",
+    "overlaps_mutation",
+    "source_sequence",
+    "offset",
+    "occurs_in_reference",
+])
+
+def logistic_epitope_score(
```
julia326: why the change to make this a method outside the EpitopePrediction object? Looks like an unrelated refactor - we should stick to one (logical) change per PR, with the exception of adding comments/fixing whitespace/etc.

vaxrank/epitope_prediction.py
```diff
@@ -109,6 +88,24 @@ def index_contains_kmer(fm, kmer):
         break
     return found
 
+def generate_protein_sequences(genome):
```
julia326: thanks for fix! Could you put this into a separate PR that just deals with the shellinford issue?

vaxrank/epitope_prediction.py
iskandr: Once `from_mhctools_binding_prediction` was moved out of the class it looked really confusing to inherit from a namedtuple just to extend it with a single helper method.

vaxrank/epitope_prediction.py
iskandr: If I test that fix alone without adjusting the mhctools code then tests will fail due to `source_sequence`. Conversely, if I fix the mhctools issue first then tests fail in Py2 due to shellinford. Is it a good idea to merge a PR with failing tests in anticipation of fixing in a second related PR? Or more generally, what do you do when fixing two bugs simultaneously?

iskandr: Shellinford fix alone as a PR: https://github.com/hammerlab/vaxrank/pull/118

julia326: thanks! can you rebase in this branch to pick up those changes?

PR #140, opened 2017-07-07
Adding an argument to specify the file path for the vaxrank Python log

### LINKED ISSUE #130: Vaxrank fails when run within a folder that the user doesn't have write access to due to logging activity
This is the configuration that is causing a problem within the epidisco setup where we, by default, call programs within the root directory and when we do that, our run fails since the user can't create/write to `/python.log`. 

Ideally, we can use a platform-independent semi-temporary file handler here; but this shouldn't fail for most use cases, so is not that urgent given that epidisco's issue can be considered an edge case.

julia326:
Defaults to python.log.

Fixes #130


vaxrank/cli.py
```diff
@@ -278,6 +281,10 @@ def main(args_list=None):
         arg_parser = new_run_arg_parser()
 
     args = arg_parser.parse_args(args_list)
+    logging.config.fileConfig(
+        pkg_resources.resource_filename(__name__, 'logging.conf'),
+            defaults={'logfilename': args.log_path})
```
armish: Glad developers thought about this use case and exposed the API. Hope it wasn't one of those hidden features that caused you trouble to get it working.

PR #141, opened 2017-07-08
Some changes in preparation for computing a minimal epitope report:

julia326:
- making EpitopePrediction a class, moving logistic_epitope_score into its methods, adding a wildtype peptide to its members (reference epitope computation logic borrowed from TESLA code, see https://github.com/hammerlab/tesla/blob/df7bfdb6241899d9b4fa2bb1dc359c2ad4935979/tesla/cli.py#L81)
- adding a predicted_effect() method to MutantProteinFragment, changing a couple places in the code to use that for computing the top priority effect of a variant given a mutant protein fragment; also adding a global_start_pos() method (see same TESLA code for reference)
- adding a hook into the CLI to make a min epitope report, not implemented yet
- setting --min-epitope-score to 0.0 in the test script so I can see multiple variants

Start of work for #142

### DIFF
--- a/vaxrank/cli.py
+++ b/vaxrank/cli.py
@@ -133,7 +133,8 @@ def add_output_args(arg_parser):
     output_args_group.add_argument(
         "--output-min-epitope-report",
         default="",
-        help="Path to the minimal epitope report")
+        help="Path to the minimal epitope report, which lists which mutant and wildtype peptides "
+        "can be used for T-cell response assays")
 
 
 def add_vaccine_peptide_args(arg_parser):
--- a/vaxrank/cli.py
+++ b/vaxrank/cli.py
@@ -37,6 +37,7 @@ from .report import (
     make_html_report,
     make_pdf_report,
     make_csv_report,
+    make_min_epitope_report,
     TemplateDataCreator,
     PatientInfo,
 )
@@ -128,6 +129,11 @@ def add_output_args(arg_parser):
         "--log-path",
         default="python.log",
         help="File path to write the vaxrank Python log to")
+
+    output_args_group.add_argument(
+        "--output-min-epitope-report",
+        default="",
+        help="Path to the minimal epitope report")
 
 
 def add_vaccine_peptide_args(arg_parser):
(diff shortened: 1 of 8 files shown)


iskandr:
Failure on Travis:

```
************* Module vaxrank.mutant_protein_fragment
E:182,12: Undefined variable 'logger' (undefined-variable)
```

vaxrank/cli.py
```diff
@@ -126,6 +127,11 @@ def add_output_args(arg_parser):
         default="",
         help="Name of final reviewer of report")
 
+    output_args_group.add_argument(
+        "--output-min-epitope-report",
+        default="",
+        help="Path to the minimal epitope report")
```
iskandr: Maybe expand the help a little more. e.g. "Path to the minimal report, lists which mutant and wildtype peptides can be used for T-cell response assays"

vaxrank/mutant_protein_fragment.py
```diff
@@ -166,3 +168,23 @@ def top_k_subsequences(
...
+
+    def predicted_effect(self):
+        effects = [self.variant.effect_on_transcript(t) for t in
+            self.supporting_reference_transcripts]
+        predicted_effect = top_priority_effect(effects)
+        return predicted_effect
+
+    def global_start_pos(self):
+        # position of mutation start relative to the full amino acid sequence
+        global_mutation_start_pos = self.predicted_effect().aa_mutation_start_offset
```
iskandr: Is there any way to use the Isovar result to get this instead?

iskandr: I think that Vaxrank has some unresolved tension around how to interpret Isovar sequences which may differ from the reference at more positions that those predicted from DNA by Varcode. If there's an adjacent somatic variant changing an additional AA -- do we count that as part of the mutation? I guess we can't because that difference could also be germline. 

I think in the long term we need Isovar to take a second VCF (of germline variants) and use that to attribute AA differences as either germline or tumor. 

In the mean time, I guess using Varcode's predicted protein pos is OK.

julia326: Makes sense after chatting offline. I think for now we can use the reference as a proxy for self, which will work correctly most of the time, but write up an explanation of where that goes wrong and keep an eye out for such cases in the future/fix it in isovar at some point.

iskandr:
LGTM

PR #143, opened 2017-07-20
Minimal epitope report

### LINKED ISSUE #142: Minimal epitope report
From a collaborator email, what we would like to have is: "column with the wt and its affinity next to the mutant min variant - also is there a way to see if there are any non-variant related wt within the SLP with high affinity that we did not account for?"

We want for each variant/vaccine peptide combination:
- each mutant epitope included in the report needs to list its WT epitope and affinity
- list all strong binders in the peptide that don't overlap the mutation

For now, add this to ASCII report, maybe to PDF report if it doesn't mess up the formatting.

julia326:
Changes to ASCII/HTML/PDF reports:
- showing WT versions of mutant peptides and their binding strengths
- in each vaccine peptide, showing strong binders that don't overlap the mutation
- manufacturability and non-mutation-overlapping-epitope sections are optional in the report, default to true, controlled by CLI flags
- simpler (and more) page breaks so that the PDF report remains printable with all the extra data

Fixes #142


vaxrank/cli.py
```diff
@@ -73,10 +74,41 @@ def cached_run_arg_parser():
...
+    manufacturability_args = arg_parser.add_mutually_exclusive_group(required=False)
+    manufacturability_args.add_argument(
+        "--include-manufacturability-in-report",
+        dest="manufacturability",
+        action="store_true")
+    manufacturability_args.add_argument(
+        "--no-manufacturability-in-report",
+        dest="manufacturability",
+        action="store_false")
+    arg_parser.set_defaults(manufacturability=True)
```
iskandr: Remind me, how does this interact with `required=True` for the group? Will the user still have to supply one of the two args, even though there's a default?

julia326: If both are left out, still defaults to true.

PR #145, opened 2017-07-31
Travis config updates

julia326:
- adding auto-deploy to PyPi, assuming version increment
- using Trusty
- updating cache config, should be a no-op except maybe successfully caching pip files now
- installing pandoc with Conda instead of add-ons; necessary for correct README md -> rst conversion
- moving pyensembl installs into a `before_script` section


timodonnell:
LGTM

PR #146, opened 2017-09-07
Small report fixes

julia326:
- Correcting for cases where WT epitopes are too short to have binding predictions (right now this crashes the report)
- Fix to report args: respect new output args in case of generating a report from a JSON file


vaxrank/cli.py
```diff
@@ -244,6 +244,7 @@ def ranked_variant_list_with_metadata(args):
     if hasattr(args, 'input_json_file'):
         with open(args.input_json_file) as f:
             data = serializable.from_json(f.read())
+            data['args'].update(vars(args))
```
iskandr: Maybe add a small comment explaining what this is doing?

PR #149, opened 2017-10-10
Adding initial Vaxrank (detailed) docs for ReadTheDocs integration.

julia326:


docs/Makefile
```diff
@@ -0,0 +1,20 @@
...
+# Put it first so that "make" without argument is like "make help".
+help:
+	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+.PHONY: help Makefile
+
+# Catch-all target: route all unknown targets to Sphinx using the new
+# "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
+%: Makefile
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
iskandr: Does this get invoked by RtD?

docs/conf.py
```diff
@@ -0,0 +1,171 @@
...
+extensions = []
+
+# Add any paths that contain templates here, relative to this directory.
+templates_path = ['_templates']
+
+# The suffix(es) of source filenames.
+# You can specify multiple suffix as a list of string:
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = '.rst'
```
iskandr: Maybe an obvious question, but why not MD?

docs/Makefile
julia326: presumably yeah; this was auto-generated using the Sphinx instructions at https://docs.readthedocs.io/en/latest/getting_started.html

docs/conf.py
julia326: that same manual, RtD docs express a preference for RST though they support both, so I went with that

PR #158, opened 2018-02-26
Adding IGV locus to the ASCII/PDF vaxrank report templates. 

### LINKED ISSUE #67: Different location text
Right now they look like  `chr2 g.36768911G>T`, but this is difficult to read and hard to copy/paste into IGV. How would `chr2:36768911 G>T` work?

julia326:
We can get a copy-pastable list of IGV loci from the resulting ASCII report by running:

`grep IGV <ASCII report path> | awk '{print $NF}'`

Fixes #67


.travis.yml
```diff
@@ -20,6 +19,7 @@ cache:
     - $HOME/.cache/pyensembl/GRCh38/ensembl87/
     - $HOME/.cache/pyensembl/GRCm38/ensembl87/
     - $HOME/.cache/vaxrank
+    - $HOME/.cache/pip
```
iskandr: Do you think this changes the behavior? If so, I want to standardize this across all the repos.

julia326: I... don't know that it does, but I want to see if it does?

PR #162, opened 2018-05-17
Using an environment variable to set where the reference peptide index lives.

julia326:


.travis.yml
```diff
@@ -42,6 +42,8 @@ before_install:
   - mkdir tmp
   - export NETMHC_BUNDLE_TMPDIR=$PWD/tmp
   - export PATH=$PATH:$NETMHC_BUNDLE_HOME/bin
+  # tells vaxrank where to install the reference peptide cache
+  - export REF_PEPTIDES_DIR=$HOME/.cache/vaxrank
```
iskandr: Can you prefix this something? Maybe `VAXRANK_REF_PEPTIDES_DIR`?

PR #170, opened 2018-09-24
Showing 5 significant figures for floats in Vaxrank reports.

julia326:
Also decreasing default PDF printing zoom so that stuff fits better onto a single page


run-vaxrank-b16-test-data.sh
```diff
@@ -6,7 +6,7 @@ vaxrank \
     --mhc-alleles H2-Kb,H2-Db \
     --mhc-epitope-lengths 8 \
     --padding-around-mutation 0 \
-    --min-epitope-score 0.0 \
+    --min-epitope-score 10e-100 \
```
iskandr: Why not 0?

julia326: Not necessary - I had it at 0 before because I wanted to have more than one result, but I've updated the test data since then so now we get 2 results back this way as well

PR #186, opened 2020-02-18
use new Isovar API but keep old logic for determining variant effects

iskandr:
This PR is an attempt to make the smallest possible change while still bringing Vaxrank over to the new Isovar API. To do this I'm leaving some redundant code in Vaxrank for e.g. determine variant effects, which can be removed in a later PR.

Also:
* cleaning up CLI code by pulling out little helper functions
* created `run_vaxrank_from_parsed_args`, which simplifies testing
* moved logic for `--version` into ArgumentParser instance

Importantly this PR allows use of newer versions of pysam (>0.9), as requested in https://github.com/openvax/vaxrank/issues/179 and by several other people offline

After the next PR I'll start actually re-running PGV patients and comparing reports.


test/test_shell_script.py
```diff
@@ -30,7 +30,8 @@
     "--mhc-predictor", "random",
     "--mhc-alleles", "H2-Kb,H2-Db",
     "--padding-around-mutation", "5",
-    "--include-mismatches-after-variant"
+    # TODO: figure out what happened to this argument in Isovar
```
julia326: ?

vaxrank/cli.py
```diff
@@ -204,30 +194,40 @@ def add_vaccine_peptide_args(arg_parser):
...
         "--min-epitope-score",
         default=1e-10,
         type=float,
         help=(
             "Ignore predicted MHC ligands whose normalized binding score "
-            "falls below this threshold"))
+            "falls below this threshold. (default: %(default)s)"))
+
+    vaccine_peptide_group.add_argument(
+        "--num-epitopes-per-vaccine-peptide",
```
julia326: why rename this arg?

vaxrank/core_logic.py
```diff
@@ -211,26 +160,29 @@ def vaccine_peptides_for_variant(
     At this point, we know the variant has RNA support, as per the
```
julia326: this comment is no longer true

vaxrank/core_logic.py
```diff
@@ -211,26 +160,29 @@ def vaccine_peptides_for_variant(
...
-    protein_fragment = MutantProteinFragment.from_isovar_protein_sequence(
-        variant=variant,
-        protein_sequence=isovar_protein_sequence)
+    if not isovar_result.passes_all_filters:
+        # don't consider candidate vaccine peptides from variants which either
+        # failed their filters or don't have an RNA-derived protein sequence
+        return []
+
+    variant = isovar_result.variant
+    long_protein_fragment = MutantProteinFragment.from_isovar_result(isovar_result)
```
julia326: why rename this to "long"? is the meaning different from before?

test/test_shell_script.py
iskandr: Found the new name, it's   --count-mismatches-after-variant

vaxrank/cli.py
iskandr: I think "per peptide" is too vague given that each epitope is itself a peptide and the longer sequence from which the vaccine peptides are drawn are also peptides.

vaxrank/core_logic.py
iskandr: Removed

vaxrank/core_logic.py
iskandr: I wanted to distinguish the 35mer source sequence from the 25mer vaccine peptides below

PR #191, opened 2020-08-17
Use Isovar for all variant counts, change Python versions in tests to 3.6 and 3.7

iskandr:
Hopefully fixes: https://github.com/openvax/vaxrank/issues/190

Also added `has_vaccine_peptide` to all passing variants alongside `mhc_binder`. They currently have identical values but eventually we could allow MHC ligand counts even for variants without RNA support. 

ALSO: one of our dependencies, BioPython, no longer supports Python 2.7 or Python 3.5. Rather than pin on an old version, I'm just removing Python 2.7 and 3.5 from the Travis test matrix and using versions 3.6 and 3.7. 

THE TIME HAS COME.


vaxrank/vaxrank_results.py
```diff
@@ -105,29 +95,32 @@ def variant_properties(self, gene_pathway_check=None):
...
-        for variant in self.variants:
-            gene_name = ''
-            if variant.gene_names:
-                gene_name = variant.effects().top_priority_effect().gene_name
+        for isovar_result in self.isovar_results:
+            variant = isovar_result.variant
+            if isovar_result.top_protein_sequence:
+                gene_name = isovar_result.top_protein_sequence.gene_name
+            else:
+                gene_name = isovar_result.overlapping_gene_names(only_coding=False)
```
julia326, line 103: this is a list, do you want to pick one or you mean to keep the whole list?

vaxrank/vaxrank_results.py
```diff
@@ -105,29 +95,32 @@ def variant_properties(self, gene_pathway_check=None):
...
                 ('contig', variant.contig),
                 ('start', variant.start),
                 ('ref', variant.ref),
                 ('alt', variant.alt),
-                ('is_coding_nonsynonymous', False),
-                ('rna_support', False),
-                ('mhc_binder', False),
+                ('is_coding_nonsynonymous',
+                    isovar_result.predicted_effect_modifies_protein_sequence),
+                ('rna_support',
```
julia326, line 112: wooo! Just checking: `rna_support` will now be true if there's RNA evidence even if there are no MHC binders, yes?

vaxrank/vaxrank_results.py
iskandr: Good catch, this supposed to be `";".join()`

vaxrank/vaxrank_results.py
iskandr: Correct

PR #296, opened 2026-05-14
test.sh: strip to 13 lines; coverage moves to CI

iskandr:
## What

``test.sh`` is now 14 lines: shebang, ``set -eo pipefail``, the macOS DYLD guard for WeasyPrint/Pango, and ``exec pytest tests "$@"``. Down from 118 lines.

CI (.github/workflows/tests.yml) calls ``pytest --cov=vaxrank/ --cov-report=term-missing tests`` directly so the Coveralls upload still gets a ``.coverage`` file. ``deploy.sh`` calls ``./test.sh`` unchanged — it no longer prints a coverage report at deploy time, but never gated on one.

## Why

Measurement on the current suite (matrix from the original PR thread):

| Run | Wall time | Δ vs serial no-cov |
|---|---|---|
| Serial, no cov | **49s** | baseline (now the default) |
| Serial + --cov | 78s | +60% (coverage instrumentation) |
| xdist -n 1, no cov | 60s | +22% (worker fork + IPC) |
| xdist -n 2, no cov | 95s | +94% — slower than -n 1 |
| Deploy: xdist -n 4 + --cov | often 20+ min | thrashes when RAM pressured |

xdist is consistently slower than serial for this suite — two structural reasons:

1. ``test_reference_peptide_logic`` runs ~20s on its own (first-time mouse kmer-set pickle load). xdist's wall-time floor IS that test; parallelism over the other 831 fast tests can't beat the single long pole.
2. Every xdist worker is a fresh Python process — each independently loads pyensembl + mhctools + topiary (~3 GB resident) and decompresses its own copy of the mouse kmer-set pickle. With N workers, N concurrent gunzip + unpickle calls hit the same file → disk contention. With ``--cov`` adding +60% on top, memory pressure pushes the kernel into swap and the deploy wedges.

Coverage in the local runner had a similar shape — useful for CI's Coveralls badge but a tax on every inner-loop run. Splitting it out of ``test.sh`` makes the default fast.

## Dropped env vars

``PER_WORKER_GB``, ``TEST_SH_MIN``, ``TEST_SH_MAX`` (added in #292) are gone. If anyone set them in a shell rc or local CI config, they silently no-op now.

## Test plan
- [x] Local ``./test.sh`` runs in ~69s, 833 passed.
- [x] Local ``./test.sh --cov=vaxrank/`` still works via the ``"$@"`` pass-through.
- [x] ``bash -n test.sh`` clean.
- [x] DYLD guard preserved for macOS WeasyPrint.
- [ ] CI run on the PR confirms the new ``pytest --cov`` invocation produces ``.coverage`` and the Coveralls step still uploads.


iskandr:
## Update — simplified further

Pushed ``1684c2e`` which strips ``test.sh`` to 13 lines: shebang, the macOS DYLD guard for WeasyPrint, and ``exec pytest --cov=vaxrank/ --cov-report=term-missing tests "$@"``.

The previous revision of this PR kept an opt-in ``TEST_SH_XDIST=1`` escape hatch. Removed in this commit — xdist is consistently slower for this suite regardless of worker count (numbers in the original PR body still stand), so the opt-in branch was dead code in practice. If/when the suite shape changes (slow test split, many genuinely-parallel slow tests land), reintroducing parallelism can be a focused PR with fresh measurements.

134 lines removed, 5 added.

(Couldn't update the PR title/body — ``gh pr edit`` is failing on a projectCards GraphQL deprecation. Latest commit message + this comment carry the corrected framing.)

PR #348, opened 2026-08-27
Make allele attribution of peptide-level evidence an explicit, recorded policy

iskandr:
Follow-up to #345. Answers the question that PR deferred: **which MHC alleles should peptide-level evidence be credited to, and how does a reader find out why?**

## The problem

Antigen processing and proteasomal cleavage describe a peptide and its flanks — no MHC involved. Vaxrank's scoring substrate is per-allele (`CandidateEpitope.per_allele_scores`), so that evidence has to be attributed to one or more alleles before it can score anything.

That choice was being made implicitly: the evidence was projected onto whichever alleles the input file happened to score the peptide against. **Row incidence is neither the patient's genotype nor a claim about presentation.** On the real LENS fixtures those sets are 1–2 alleles out of a 5–6 allele genotype, so the implicit answer was an artifact of how the upstream tool chose to write its rows.

## The policy

`EpitopeConfig.allele_free_evidence`:

| value | meaning |
|---|---|
| `selected` *(default)* | the allele that ranks best for this peptide; falls back to the whole genotype when the peptide has no allele-scoped evidence to rank on |
| `all` | every allele in the patient's genotype |
| `top:N` | the N best, same fallback |
| `from_input` | only pairings the source file already carried — pre-3.3 behavior |

The fallback is the point of the default: *"we don't know which allele presents this"* is not the same claim as *"this one does."* When the model can pick, it does; when it can't, the evidence is preserved across the genotype rather than assigned to an arbitrary winner.

`allele_selection_axis` chooses the ranking axis. `auto` prefers presentation when the input carries it, else affinity — presentation being the model's own statement about the question actually being asked.

## Where the allele comes from, and how to reconstruct it

This is the part worth reviewing closely. Every attributed allele carries an `AlleleAttribution` recording **why**:

- `source` — **how the evidence reached this allele**, one of:
  - `from_input` — the source file already paired this peptide with this allele
  - `broadcast` — nothing narrowed it, so every genotype allele received it
  - `selected` — vaxrank ranked the genotype and picked this one
- for a selection: `rank_kind`, `rank_predictor`, `rank_value`, `rank_position`

The names describe the *mechanism*, not where the allele sits. An earlier draft used `observed`/`genotype`; both were wrong in the same way. `genotype` distinguished nothing — every allele here is a genotype allele — and `observed` implied empirical evidence of presentation when it only meant a row existed in the file, which is an artifact of what the upstream tool chose to write.

So a finished run says, in full:

```
HLA-A*02:01 (selected: ranked #1 on pMHC_presentation [mhcflurry] = 0.8)
HLA-C*07:01 (broadcast: no allele-specific evidence)
HLA-B*07:02 (from the input file)
```

Two invariants enforced in `__post_init__`: a selection without a recorded axis is rejected (it could not be reconstructed), and ranking provenance on a non-selection is rejected (it would imply a ranking that never happened).

These are **serialized in the native format**. That is not incidental — the selection depends on the config in force at predict time, so a reload under a different config would otherwise silently re-derive a different answer, which is the failure mode #345 spent its length removing.

The genotype itself resolves through `frame_genotype()`: an explicit genotype wins, and absent one it falls back to the union of alleles the inputs scored — the widest honest statement the data supports.

## Impact, measured

On the real LENS fixtures with `score_expr=processing[mhcflurry].score`:

| fixture | policy | allele groups | summed score | top-10 churn |
|---|---|---|---|---|
| v1.4 | from_input | 30 | 4.676 | — |
| v1.4 | selected | 30 | 4.676 | 0/10 |
| v1.4 | **all** | **146** | **23.380** | 0/10 |
| v1.9+stops | from_input | 52 | 12.930 | — |
| v1.9+stops | selected | 52 | 12.041 | **6/10** |
| v1.9+stops | **all** | **245** | **60.204** | 6/10 |

`all` multiplies the summed score 4.7–5.0× by crediting peptide-level evidence to alleles the input never scored — which is why it is not the default, despite being the most "complete" reading. `selected` keeps the same allele groups as today and moves 6 of the top 10 on one fixture.

**Runs on the default `score_expr` are unaffected by any of this** — it is `(affinity.value < 5000.0) * affinity.value.logistic_normalized(350.0, 150.0)`, which references affinity only, and the default filter is `None`. Only configs that explicitly reference `processing` or `cleavage` see a change, which is an opt-in population.

## Why this is not delegated to topiary

topiary 5.20.0's `peptide_view()` and `alleles=` were expected to let this projection be deleted. `alleles=` declares **one allele set for a whole frame**, and every policy except `all` needs a different set per peptide — verified against the published wheel. So the projection stays; it is load-bearing policy rather than a workaround, and it is now named as such.

topiary's floor is unchanged at `>=5.17.1`.

## Also

Closes the two branch gaps in `_topiary_pvacseq_to_epitope_rows` that no fixture had ever taken: a pVACseq row with no WT columns (topiary leaves `wt_value` NA — the shape an aggregate written without WT prediction produces) and a row with no IC50 at all. Both are real input shapes, neither had ever executed, so nothing said what they should do. Now: the WT-less row contributes no WT leaf rather than a fabricated zero, and the unscored row is dropped rather than carried.

Branch coverage for that function goes from two open branches to none.

## Verification

1045 passed, 1 skipped, ruff clean, on topiary 5.20.0. Version 3.2.0 → 3.3.0.

### DIFF
--- a/tests/test_allele_evidence.py
+++ b/tests/test_allele_evidence.py
@@ -212,3 +212,96 @@ def test_attribution_requires_reconstructable_provenance():
         AlleleAttribution(
             allele="HLA-A*02:01", source=ALLELE_SOURCE_BROADCAST,
             rank_kind="pMHC_affinity")
+
+
+def test_predictor_is_chosen_by_coverage_unless_the_config_names_one():
+    """A predictor covering one allele must not silently reduce a top:3.
+
+    Preference order alone picked the canonical name even when it could rank
+    a single allele out of six — the same silent reduction as choosing the
+    axis by presence rather than coverage, arriving through a second door.
+    """
+    genotype = ["HLA-A*02:01", "HLA-B*07:02", "HLA-C*07:01"]
+    epitope = _epitope(
+        _affinity("HLA-A*02:01", 50.0, "mhcflurry"),      # covers 1
+        _affinity("HLA-A*02:01", 900.0, "netmhcpan"),     # covers 3
+        _affinity("HLA-B*07:02", 10.0, "netmhcpan"),
+        _affinity("HLA-C*07:01", 20.0, "netmhcpan"),
+        _processing(),
+        patient_alleles=tuple(genotype))
+
+    got = attribute_alleles(epitope, genotype, AllelePolicy.parse("top:3"))
+    assert [a.allele for a in got] == [
+        "HLA-B*07:02", "HLA-C*07:01", "HLA-A*02:01"]
+    assert {a.rank_predictor for a in got} == {"netmhcpan"}
+
+    # An explicit choice is the user's, and still wins over coverage.
+    named = attribute_alleles(
+        epitope, genotype, AllelePolicy.parse("top:3"),
+        default_methods={"pMHC_affinity": "mhcflurry"})
+    assert [(a.allele, a.rank_predictor) for a in named] == [
+        ("HLA-A*02:01", "mhcflurry")]
+
+
+def test_recorded_values_survive_serialization():
+    """Predictions from a pandas frame carry numpy scalars.
+
+    A recorded value has to round-trip to stay reconstructable, and the
+    serializer refuses numpy types outright, so the coercion happens where
+    the value is captured.
+    """
+    import numpy as np
+
+    from vaxrank.allele_evidence import AlleleAttribution
+
+    epitope = _epitope(
+        _affinity("HLA-A*02:01", np.float64(50.0)), _processing(),
+        patient_alleles=("HLA-A*02:01",))
+    [attribution] = attribute_alleles(
+        epitope, ["HLA-A*02:01"], AllelePolicy.parse("selected"))
+
+    assert type(attribution.rank_value) is float
+    assert AlleleAttribution.from_json(attribution.to_json()) == attribution
+
+
+def test_mhc_dependence_table_covers_every_topiary_kind():
+    """A new prediction kind must be classified, not silently un-attributable.
+
+    "Is this prediction allele-scoped?" used to be re-derived inline at four
+    call sites, two of them testing only for a blank allele — which is also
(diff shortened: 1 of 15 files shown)


iskandr:
Superseded by #371, #372 and #373, which shipped this feature with the review findings addressed rather than rebasing eight commits onto a design that had moved.

What landed:

- **#371** — attribution at the frame layer, so both input paths honor the policy (closes #349). This branch attributed on `CandidateEpitope` objects, which only the external report path builds before scoring, so the setting applied to LENS runs and silently did nothing on the native pipeline. Also closes #357: MHC dependence is read from `topiary.KIND_MHC_DEPENDENCE` rather than the local partition this branch carried, which named one peptide-level kind where topiary names five.
- **#372** — the attribution round-trips through the native format.
- **#373** — it appears in the neoepitope report, so a reader can tell which allele a peptide's processing score is credited to.

Three changes from what this branch proposed, all from its own review:

1. **The default is `all`, not `selected`.** `all` is what vaxrank has always done, so the series changes no scores. Making `selected` the default is a semantic decision with evidence to gather, and is #351.
2. **The knob is reachable.** It was settable only by constructing `EpitopeConfig` in Python — no YAML, no CLI.
3. **Narrowing does not yet change scoring, and says so.** topiary projects a peptide-level value across every allele group keyed on the *kind*, so a row that names an allele is projected onto the others anyway — openvax/topiary#232. Writing rows onto the attributed alleles was implemented, measured and removed: the frame changed and the scores did not, which is worse than the gap. A narrowing policy warns once, naming the issue.

The vocabulary from this branch survived intact — `ALLELE_SOURCE_FROM_INPUT` / `_BROADCAST` / `_SELECTED`, the policy names, and `RankedEvidence` keeping kind, predictor and values in one object so a mismatched `(predictor, value)` pair is unrepresentable.

Issue #351, opened 2026-08-28
Decide two allele-attribution semantics: short top:N, and where the genotype comes from


iskandr:
Two decisions in #348's allele attribution are currently defaults-by-implementation rather than stated policy. Both are defensible as they stand; neither has been decided on purpose.

## 1. `top:N` when fewer than N alleles are rankable

An axis ranks only the alleles that carry evidence on it. With a three-allele genotype where one allele has affinity data, `top:3` returns **one** attribution; the other two get nothing.

Currently documented as "up to N of the alleles the axis could rank", and the alternative — padding with `broadcast` attributions for the unrankable alleles — was rejected because it asserts a preference the data does not support. But the stated rule for the default policy is *"if we don't know which allele presents this, preserve them all"*, and this case is precisely "we don't know" for the unranked alleles. The two readings genuinely conflict.

Options: keep (short result), pad with `broadcast`, or make it configurable.

## 2. Where the genotype comes from

`frame_genotype` falls back to the union of `patient_alleles` across the frame — the alleles the input scored *somewhere*. That is a proxy for the patient's genotype, not the genotype.

vaxrank already has a better source on the external path: `patient_info_from_external` infers the allele set and tags it `(inferred from report)`, and the pipeline path has `--mhc-alleles`. Neither is connected to attribution. A patient typed for six alleles whose LENS file only ever scored four would broadcast to four.

Fix: thread the resolved patient genotype (explicit `--mhc-alleles`, else the inferred set) into scoring, and keep the frame union only as the last resort it was meant to be. Note `attribute_alleles` already honors an explicit genotype outright — the plumbing to supply one is what is missing.

Found while auditing #348.

PR #371, opened 2026-08-31
Attribute alleles at the frame layer so both input paths agree

### LINKED ISSUE #349: allele_free_evidence is a no-op on the native VCF pipeline
`EpitopeConfig.allele_free_evidence` (added in #348) is honored only where the evaluation frame is built by `epitope_dsl.epitopes_to_topiary_df`. The native VCF/BAM pipeline does not use it, so the knob silently does nothing on the path most runs take.

## Mechanism

`epitope_logic.predict_epitopes` gets a frame straight from the predictor and scores it directly:

```python
predictions_df = topiary_predictor.predict_from_named_sequences(...)
predictions_df = apply_filter(predictions_df, filter_node, group_keys=..., ...)
score_ctx = EvalContext(predictions_df, group_keys=..., ...)
```

`epitopes_to_topiary_df` — the only place attribution runs — is never called. A peptide-level leaf (mhcflurry's integrated model emits `antigen_processing`) therefore forms its own empty-allele group, contributes to no allele's score, and the resulting `CandidateEpitope` carries `allele_attributions = ()`.

## Why it matters

This is the same failure class #345 removed for `vaccine_peptides:` config — a setting that applies on one input path and quietly does nothing on another. It is worse here because the field is documented on `EpitopeConfig` as a general knob and 3.3.0 changes its default, so a user reading the config has no way to tell the setting is inert for their run.

It also means the feature currently works on the LENS path alone. pVACseq supplies its own topiary frame (so attribution is deliberately skipped, correctly, since topiary's pVACseq reader emits no peptide-level kinds), and the native pipeline never reaches it.

## Fix

Attribution is implemented one layer too shallow — on `CandidateEpitope` objects, when the thing both paths share is a **topiary frame**. Move it: a frame-level step that takes `(frame, policy, genotype, default_methods)` and returns the frame with peptide-level rows attributed to alleles, called by both `predict_epitopes` and `epitopes_to_topiary_df`.

That requires `attribute_alleles` to rank from a group of frame rows rather than from a `CandidateEpitope`'s `predictions_flat()`. The ranking logic itself is unchanged — only its input shape.

Until that lands, #348 should refuse or warn when a non-default policy is set on a path that ignores it, rather than accepting it silently.

Found by code review of #348.

### LINKED ISSUE #357: Replace the local MHC-dependence table with topiary's public KIND_MHC_DEPENDENCE
`vaxrank/allele_evidence.py` maintains its own classification of prediction kinds as peptide-level or allele-scoped (`PEPTIDE_LEVEL_KINDS` / `ALLELE_SCOPED_KINDS`). topiary 5.22.0 makes the same fact public, so the copy can go.

```python
from topiary.ranking import KIND_MHC_DEPENDENCE, mhc_dependence, MHC_DEPENDENCE_VALUES

KIND_MHC_DEPENDENCE["antigen_processing"]  # "none"
KIND_MHC_DEPENDENCE["pMHC_affinity"]       # "single_allele"
```

Filed upstream as openvax/topiary#195, delivered in 5.22.0. `KIND_MHC_DEPENDENCE` is a `MappingProxyType`, so an unknown key raises `KeyError` rather than defaulting, and `MHC_DEPENDENCE_VALUES` is re-exported from mhctools rather than hand-copied.

## The two tables agree today

Checked against the published 5.22.0 wheel, all ten canonical kinds:

| kind | topiary | vaxrank |
|---|---|---|
| `antigen_processing` | `none` | peptide-level |
| `endolysosomal_cleavage` | `none` | peptide-level |
| `erap_trimming` | `none` | peptide-level |
| `proteasome_cleavage` | `none` | peptide-level |
| `tap_transport` | `none` | peptide-level |
| `immunogenicity` | `single_allele` | allele-scoped |
| `pMHC_TCR_binding` | `single_allele` | allele-scoped |
| `pMHC_affinity` | `single_allele` | allele-scoped |
| `pMHC_presentation` | `single_allele` | allele-scoped |
| `pMHC_stability` | `single_allele` | allele-scoped |

**No disagreements.** So this is a pure de-duplication with no behavior change — worth doing because nothing currently detects the two drifting apart in *classification* (the existing CI test only checks *completeness* against `KIND_ALIASES`), not because the table is wrong.

## Cost

Requires raising `requirements.txt` from `topiary>=5.17.1` to `>=5.22.0`, which forces every user onto a newer topiary. That is the reason this is not folded into #348 — it is a packaging decision that should be visible on its own, not bundled into a PR already carrying a scoring-behavior change.

## Scope

- Replace the two frozensets and `is_peptide_level` / `is_allele_scoped` internals with lookups against `KIND_MHC_DEPENDENCE`.
- Keep `is_peptide_level` requiring **both** halves — the kind carries no MHC *and* the record is actually allele-free. topiary's table answers only the first half; the second is what separates a correctly allele-free record from a malformed allele-scoped one, which is the distinction that caused the defect in #348.
- Drop `unclassified_kinds()` and its import-time warning; `KIND_MHC_DEPENDENCE` raising `KeyError` on an unknown kind covers it.
- Note one nuance from upstream: the table's value for `pMHC_presentation` is `single_allele`, because haplotype is a per-run mode rather than a property of the kind. `"haplotype"` only ever arrives via `kind_support` or an `allele_set`, never from the table. Harmless for the none/not-none split vaxrank needs, but it means the table alone is not a complete dependence answer.

iskandr:
Closes #349. Also closes #357.

`allele_free_evidence` was implemented on `CandidateEpitope` objects, which only the external report path builds before scoring. The native VCF/BAM pipeline scores a frame straight from the predictor and never reached the code — so the setting applied to LENS runs and **silently did nothing on the path most runs take**.

The thing both paths share is a topiary frame, so attribution moves there. `attribute_frame(frame, policy, genotype, default_methods)` returns attributions keyed by the peptide identity both paths group on, and is called from `score_predictions` and from `predict_epitopes`. A parameterized test runs every policy against both frame shapes and asserts they agree — the property that was missing.

The native path passes **no** genotype: it predicts every peptide against the patient's full typing, so the alleles the frame already names *are* the genotype. Deriving it beats threading a second, possibly disagreeing, copy through.

### Three fixes from the review of the first attempt

- **The knob is reachable.** It was settable only by constructing `EpitopeConfig` in Python — no YAML, no CLI. Both now carry it, with a test.
- **The default is `all`**, what vaxrank has always done, so this changes no scores. Whether it should become `selected` is a semantic decision with evidence to gather: #351.
- **MHC dependence comes from `topiary.KIND_MHC_DEPENDENCE`** rather than a local copy (#357). The copy named one peptide-level kind where topiary names five.

### What this does not do — and says so at runtime

A narrowing policy **does not change per-allele scores**. topiary projects a peptide-level value across every allele group keyed on the *kind*, so a row that names an allele is projected onto the others anyway:

```
row: kind=antigen_processing, allele=HLA-A*02:01, score=0.8

antigen_processing['mhcflurry'].score
  HLA-A*02:01    0.8
  HLA-B*07:02    0.8      <- the row said A*02:01
```

Filed as openvax/topiary#232.

Writing rows onto the attributed alleles was implemented, measured, and **removed**: the frame changed and the scores did not, which is worse than the gap — the frame then states something the evaluation discards. A narrowing policy now warns once, naming the issue and saying scores currently match `all`.

So `selected` today means "record and report which allele we credit", not "score only that allele". That is a real, visible effect (and what #348 surfaces), but it is not the whole feature, and the warning says which half is missing rather than letting a user infer it.

### Verification

- All eight LENS fixtures **byte-identical**.
- 1138 tests pass, including 24 new ones.

### For review

The open question is the one above: is "records the credited allele, does not yet narrow scoring" the right thing to ship, or should narrowing policies refuse outright until topiary#232 lands? I went with warn-and-record because the record is independently useful and #348 builds on it.

### DIFF
--- a/tests/test_allele_evidence.py
+++ b/tests/test_allele_evidence.py
@@ -1,0 +1,205 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for frame-level allele attribution."""
+
+import pandas as pd
+import pytest
+
+from vaxrank.allele_evidence import (
+    ALLELE_SOURCE_BROADCAST,
+    ALLELE_SOURCE_FROM_INPUT,
+    ALLELE_SOURCE_SELECTED,
+    AllelePolicy,
+    AllelePolicyError,
+    attribute_frame,
+    is_peptide_level_kind,
+)
+from vaxrank.epitope_dsl import (
+    PIPELINE_GROUP_COLUMNS,
+    PREDICTION_GROUP_COLUMNS,
+)
+
+A = "HLA-A*02:01"
+B = "HLA-B*07:02"
+
+
+def _frame(identity_column, *, processing=True):
+    """One peptide: strong affinity for A, weak for B, plus a processing row."""
+    rows = [
+        {identity_column: "s", "peptide": "SIINFEKL", "peptide_offset": 0,
+         "allele": allele, "kind": "pMHC_affinity",
+         "prediction_method_name": "mhcflurry", "predictor_version": "2.1.1",
+         "value": affinity, "score": None, "percentile_rank": None}
+        for allele, affinity in ((A, 50.0), (B, 900.0))]
+    if processing:
+        rows.append(
+            {identity_column: "s", "peptide": "SIINFEKL", "peptide_offset": 0,
+             "allele": "", "kind": "antigen_processing",
+             "prediction_method_name": "mhcflurry",
+             "predictor_version": "2.1.1", "value": None, "score": 0.8,
+             "percentile_rank": None})
+    return pd.DataFrame(rows)
+
+
+FRAME_SHAPES = [
+    pytest.param("prediction_id", list(PREDICTION_GROUP_COLUMNS), id="external"),
+    pytest.param(
+        "source_sequence_name", list(PIPELINE_GROUP_COLUMNS), id="native"),
+]
+
+
+@pytest.mark.parametrize("identity_column,group_columns", FRAME_SHAPES)
+@pytest.mark.parametrize("policy_text,expected", [
+    ("all", [A, B]),
+    ("selected", [A]),
+    ("top:2", [A, B]),
+    ("top:1", [A]),
+    ("from_input", [A, B]),
+])
+def test_both_input_paths_attribute_identically(
+        identity_column, group_columns, policy_text, expected):
+    """The same policy must produce the same answer on either frame shape.
+
+    This is the point of attributing at the frame layer. The previous
+    implementation ran on CandidateEpitope objects, which only the external
+    report path builds before scoring, so the setting applied to LENS runs
+    and silently did nothing on the native pipeline — the path most runs
+    take (openvax/vaxrank#349).
+    """
+    policy = AllelePolicy.parse(policy_text)
+    attributions = attribute_frame(
+        _frame(identity_column), policy, group_columns=group_columns)
+
+    [attributed] = attributions.values()
+    assert [a.allele for a in attributed] == expected
+
+
+@pytest.mark.parametrize("identity_column,group_columns", FRAME_SHAPES)
+def test_a_selection_records_how_it_was_made(identity_column, group_columns):
+    """A selected allele carries the evidence that selected it.
+
+    Without this the attribution is an assertion the reader cannot check.
+    """
+    attributions = attribute_frame(
+        _frame(identity_column), AllelePolicy.parse("selected"),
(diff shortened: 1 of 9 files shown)


iskandr:
Reviewed this and found two problems, both the same thing: **the attribution was computed and then not used.**

### It ran where nothing consumed it

`score_predictions` computed a full attribution mapping and used it only to decide whether to warn. Attribution belongs where candidates are built, not where scores are computed — so it moved to `attach_per_allele_scores` and to the native pipeline's row assembly, both of which now keep it on a new `CandidateEpitope.allele_attributions` field.

That also makes the knob observable, which it wasn't before:

```
all         A*02:01 (from the input file), B*07:02 (from the input file)
selected    A*02:01 (selected: ranked #1 on pMHC_affinity [mhcflurry] = 50.0)
top:2       A*02:01 (#1 ... = 50.0), B*07:02 (#2 ... = 900.0)
```

### It walked every row to discover there was nothing to do

Most frames carry no allele-free evidence, and `df.to_dict("records")` costs **0.43s on a 200k-row frame** — paid on every scoring pass, for an empty result. There's a vectorized pre-check now, and a test asserting the frame is never materialized when there's nothing to attribute.

`attach_per_allele_scores` also builds the frame once and shares it with `score_predictions`, instead of building it twice.

### One more test

Per-allele scores are pinned identical between `selected` and `all`, so the warning's claim about openvax/topiary#232 fails when it stops being true rather than quietly going stale.

PR #372, opened 2026-08-31
Round-trip allele attributions through the native format

iskandr:
#371 records which alleles were credited with a candidate's allele-free evidence, and why. The native format dropped it, so a saved run reloaded without the answer it had made.

### Why it is stored rather than recomputed

The attribution depends on the policy in force when the candidate was **scored**, and `load_predictions` takes no config. A run scored under `selected` credits one allele; reloading it would quietly produce the two-allele answer the default gives — and present that as what the run produced.

The whole value of the record is explaining a finished ranking later, which needs the run's own answer, not a fresh derivation of a different question. Same property, and same reason, as `patient_alleles` and `per_allele_scores`, which it is serialized alongside.

### Verification

```
selected  round-trips=True  ['HLA-A*02:01 (selected: ranked #1 on pMHC_affinity [mhcflurry] = 50.0)']
top:2     round-trips=True  ['HLA-A*02:01 (#1 ... = 50.0)', 'HLA-B*07:02 (#2 ... = 900.0)']
all       round-trips=True  ['HLA-A*02:01 (from the input file)', 'HLA-B*07:02 (from the input file)']
```

Tests cover every policy and assert the **ranking provenance** survives, not just the allele name — an attribution without the axis, predictor and value that produced it is an assertion the reader cannot check.

- All eight LENS fixtures byte-identical.
- 1145 tests pass.

### DIFF
--- a/tests/test_allele_evidence.py
+++ b/tests/test_allele_evidence.py
@@ -262,3 +262,70 @@ def test_a_frame_with_no_allele_free_evidence_is_not_walked():
         counting, AllelePolicy.parse("selected"),
         group_columns=list(PREDICTION_GROUP_COLUMNS)) == {}
     assert calls == []
+
+
+@pytest.mark.parametrize("policy_text", ["all", "selected", "top:2", "from_input"])
+def test_attributions_survive_a_native_round_trip(tmp_path, policy_text):
+    """A saved run must reload with the attribution it actually made.
+
+    The attribution depends on the policy in force when the candidate was
+    scored, and ``load_predictions`` takes no config — so recomputing on
+    load would silently answer a different question and present it as what
+    the run produced. Recording it is what makes a finished ranking
+    explainable later.
+    """
+    from vaxrank.epitope_config import EpitopeConfig
+    from vaxrank.epitope_io import (
+        load_predictions, read_lens_report, save_predictions)
+
+    source = tmp_path / "processing.tsv"
+    source.write_text(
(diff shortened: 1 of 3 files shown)


iskandr:
Reviewed. Two things checked, one gap found.

### Backward compatibility — verified

A file written before this column existed must still load. Both shapes work:

```
column absent   -> attributions = ()
column blank    -> attributions = ()
```

`row.get("allele_attributions", "")` handles the missing column and `string_or_empty` handles the blank cell, so an old native CSV loads with an empty record rather than failing. A malformed *value* still raises, which is right — a corrupt file should be loud.

### Serialized values are builtins

`rank_value` and `rank_position` are typed `object` on the dataclass, and numpy scalars survive arithmetic unchanged — `serializable` fails on those with `Cannot convert 1 : numpy.int64`. They are coerced at construction, and the round-trip tests would fail if that regressed.

### Gap: it is not in the human-readable report

The attribution round-trips through the native format and is on the object, so programmatic consumers can read it. But the neoepitope report — the thing a person actually looks at — does not show it, and that report already prints `mhcflurry processing score` on **every** allele row, which is exactly the broadcast the policy is about. A reader currently sees the same processing score on both alleles with nothing saying which one vaxrank credits.

Not fixed here: `_build_lens_report_row` assembles from raw report rows and has no access to the candidate, so threading attributions in is its own change rather than a line in this one. Doing it as the next PR.

PR #373, opened 2026-08-31
Say in the report which allele the processing score counts for

iskandr:
The neoepitope report prints a peptide's allele-free evidence — the processing score — on **every** allele row, because that is what an allele-free value is. Under a narrowing policy only one of those alleles is credited with it, and the two rows looked identical.

```
before                        after (allele_free_evidence=selected)
HLA-A*02:01  processing 0.8   HLA-A*02:01  0.8  selected: ranked #1 on
                                                pMHC_affinity [mhcflurry] = 50.0
HLA-B*07:02  processing 0.8   HLA-B*07:02  0.8  no
```

So the attribution recorded in #371 and round-tripped in #372 is now visible to the person reading the output, not just to a programmatic consumer. The **reason** travels with it — axis, predictor, value — because an attribution the reader cannot check is an assertion.

### Notes

- Added *after* attribution rather than in the row builder: report rows are assembled while the file is being read, before any scoring or attribution has happened, so the builder cannot know the answer.
- The column is **absent** when a file carries no allele-free evidence — matching how the per-predictor value columns already behave (present when the input has that signal), rather than adding a column of "n/a" to every report.
- `report_df.attrs` survives the annotation, with a test. pVACseq stores its own topiary frame there and the scorer reads it back; losing it would silently rebuild from CandidateEpitopes and drop the passthrough annotation columns that only exist on the stored frame.

### Verification

Measured across all eight LENS fixtures: **exactly one column added** on the four that carry processing evidence, nothing removed, and no other recorded field differs anywhere — same epitopes, offsets, leaves, per-allele scores.

1149 tests pass.

### DIFF
--- a/tests/test_allele_evidence.py
+++ b/tests/test_allele_evidence.py
@@ -329,3 +329,97 @@ def test_a_reload_does_not_re_derive_under_the_default_policy(tmp_path):
     # And the ranking provenance is intact, not just the allele name.
     assert reloaded.allele_attributions[0].rank_predictor == "mhcflurry"
     assert reloaded.allele_attributions[0].rank_value == 50.0
+
+
+def _processing_report(tmp_path, policy_text):
+    from vaxrank.epitope_config import EpitopeConfig
+    from vaxrank.epitope_io import read_lens_report
+
+    source = tmp_path / "processing.tsv"
+    source.write_text(
+        "peptide\tallele\tpep_context\tmhcflurry_2.1.1.aff\t"
+        "mhcflurry_2.1.1.proc_score\n"
+        "SIINFEKL\tHLA-A02:01\tXXSIINFEKLXX\t50\t0.8\n"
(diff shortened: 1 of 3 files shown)


iskandr:
Reviewed. One real fragility, fixed.

`annotate_credited_alleles` filled the column by slicing the allele back off the front of `describe()`:

```python
return entry.describe().split(" ", 1)[1].strip("()")
```

The report already has an `Allele` column, so stripping it is the right intent — but string surgery on a human-readable sentence breaks silently the first time the sentence changes. Nothing would have failed; the column would just start carrying the wrong text.

`reason()` now returns the explanation without the allele, and `describe()` composes from it, so the two cannot drift.

Also checked: `iterrows()` over the report is the slow path here, but the report is one row per (peptide, allele) that passed a threshold — thousands, not millions — and the annotation runs once per load rather than per scoring pass. Left as is rather than vectorizing a string-formatting step.

PR #376, opened 2026-09-01
Carry how the read counts were obtained, not just the counts

### LINKED ISSUE #375: Read counts arrive with their derivation named, and vaxrank drops it
#374 fixed *which* number `n_alt_reads` is. It did not carry across *how that number was obtained*, and topiary labels it per field.

`_read_counts_from_lens_row` reads four integers off the row and discards the rest. Verified present on the rows vaxrank already holds:

```
record.row has read_count_method          True
record.row has sequence_source            True
record.row has variant_allele_expression  True
```

None of them reach `MutantProteinFragment`, and none reach the report.

## Why it matters

After #374, `n_alt_reads = 51` on a LENS row means **depth × VAF, rounded** — an estimate — while on another source it may be a genuine count from an alignment. The report prints the same integer either way, and `sqrt(n_alt_reads) * target_epitope_score` weights them identically.

topiary's vocabulary distinguishes exactly this:

| method | meaning |
|---|---|
| `rna_reads` | counted from an alignment |
| `rna_depth_x_vaf` | depth × VAF, rounded — not counted |
| `cds_overlap_reads` | counted, but of CDS overlap rather than variant support |
| `tpm_x_dna_vaf` | abundance × DNA VAF — an expression proxy |

This is the same gap #373 closed for allele attribution: the number was on the page and the reason it was that number was not, so a reader could not tell a measurement from an estimate.

## Also dropped

**`sequence_source`** — `lens_pep_context`, `pvacseq_epitope`, `varcode_translation`, `isovar_assembly`, `caller_supplied`. Per-row "assembled from RNA versus translated from reference", which is the first thing worth knowing when auditing a ranking and is currently answerable only at whole-file granularity.

**`variant_allele_expression`** — transcript abundance × DNA VAF, in its own column rather than blended into `transcript_expression`. It assumes both alleles are transcribed equally, so a variant on a transcriptionally silenced allele looks expressed. Useful as a labelled proxy, dangerous unlabelled — which is why it should arrive with its label or not at all.

## Fix

1. Add `read_count_method` and `sequence_source` to `MutantProteinFragment`, defaulting to `""` so nothing existing changes shape.
2. Populate them from the LENS row alongside the counts already read there.
3. Surface both in the neoepitope report, the way #373 surfaced the attribution reason.

No topiary dependency — the columns already reach vaxrank's rows. I previously said this was blocked on the `ProteinFragment` builder work upstream; that was wrong. What is blocked upstream is the *native* pipeline's fragments carrying these, since those are built from varcode/isovar rather than from a frame.

iskandr:
Closes #375.

#374 fixed *which* number `n_alt_reads` is. It did not carry across *how that number was obtained*, and topiary labels it per field.

After #374 a LENS `n_alt_reads` of 51 means **depth × VAF, rounded** — an estimate. On another source the same field is counted from an alignment. The report prints the same integer either way, and the combined-score DSL weights them identically: `sqrt(n_alt_reads)` does not know that 51 was arithmetic rather than observation.

### What changes

`read_count_method` and `sequence_source` are now fields on `MutantProteinFragment`, populated from the LENS row and shown in the report:

```
n_alt=51  method='rna_depth_x_vaf'  source='lens_pep_context'
n_alt=0   method=''                 source='lens_pep_context'
```

The second line is the distinction worth having. A row with no VAF gets **no estimate and no method**, rather than a zero labelled `rna_depth_x_vaf` — which would assert that the estimate was made and came out zero.

### The label travels with the counts

From the same row. The counts already come from one best-supported observation rather than per-field maxima; reading the label separately would reintroduce that problem one field over — a method describing a row whose numbers were not the ones used. A test pins it with a deep unlabelled row and a shallow labelled one.

### Scope

Both fields default to `""` so nothing existing changes shape, and the report omits the rows when a source did not say.

**Not covered:** the native VCF/BAM pipeline builds fragments from varcode and isovar rather than from a frame, so its fragments carry no label yet. That half genuinely does wait on openvax/topiary#102 — unlike the LENS half, which I had wrongly assumed was blocked upstream when the columns were already arriving on the rows.

1154 tests pass.

### DIFF
--- a/tests/test_external_input.py
+++ b/tests/test_external_input.py
@@ -1975,3 +1975,42 @@ def test_the_method_describes_the_row_whose_counts_were_used(tmp_path):
     # The winning row is the shallow one, so its depth and its label.
     assert (fragment.n_overlapping_reads, fragment.n_alt_reads) == (10, 9)
     assert fragment.read_count_method == "rna_depth_x_vaf"
+
+
+def test_pvacseq_estimates_are_labelled_too(tmp_path):
+    """pVACseq reports a depth and a fraction, never an alt-read count.
+
+    vaxrank multiplies them, which is the same estimate the LENS path
+    makes — so it carries the same label. It was computing it inline and
+    unlabelled, which is the bug #375 describes, one path over.
+    """
+    from vaxrank.epitope_config import EpitopeConfig
+    from vaxrank.epitope_dsl import epitopes_for_ranking
+    from vaxrank.epitope_io import read_pvacseq_report
+    from vaxrank.external_input import pvacseq_ranking_result
+
+    config = EpitopeConfig()
+    report = read_pvacseq_report(
+        os.path.join(DATA_DIR, "pvacseq_example.tsv"), epitope_config=config)
+    result = pvacseq_ranking_result(
+        report, epitopes_for_ranking(list(report.epitopes), config))
+
+    fragment = result.ranked[0][1][0].mutant_protein_fragment
+
+    assert fragment.n_alt_reads > 0
+    assert fragment.read_count_method == "rna_depth_x_vaf"
+
+
+def test_a_pvacseq_row_with_no_rna_vaf_claims_no_derivation():
+    """Absent is not zero.
+
+    ``pvacseq_rna_vaf`` used to default to 0.0, which made "the variant was
+    looked for and not seen" indistinguishable from "no RNA data" — and
+    would have labelled the second one ``rna_depth_x_vaf``, asserting an
+    estimate that was never made.
+    """
+    from vaxrank.external_input import pvacseq_rna_vaf
+
+    assert pvacseq_rna_vaf({"rna_vaf": 0.0}) == 0.0
+    assert pvacseq_rna_vaf({}) is None
+    assert pvacseq_rna_vaf({"rna_vaf": ""}) is None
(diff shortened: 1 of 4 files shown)


iskandr:
Reviewed. Found the same bug one path over.

### pVACseq was making the same estimate, unlabelled

pVACseq reports a **depth** and a **fraction**, never an alt-read count, so vaxrank multiplies them:

```python
counts.append((depth, int(round(depth * pvacseq_rna_vaf(record.row)))))
```

That is `rna_depth_x_vaf`, computed inline and shipped without the label — exactly what #375 describes, on the other external path. It now carries the label, from the same row the counts came from.

```
pvacseq n_alt=68   method='rna_depth_x_vaf'
```

`sequence_source` stays blank there because topiary's pVACseq reader does not emit it for these fixtures — blank rather than a guess.

### `pvacseq_rna_vaf` defaulted to 0.0

Which made *"the variant was looked for and not seen"* indistinguishable from *"there is no RNA data"* — and would have labelled the second one `rna_depth_x_vaf`, asserting an estimate nobody made. It returns `None` now, and `0.0` still reads as `0.0`:

```
rna_vaf=0.0  -> 0.0
absent       -> None
blank        -> None
```

The helper had no other caller, so this replaced its body rather than adding a second reader beside it — which is what my first cut did, and is the shape this whole thread has been about.

Issue #383, opened 2026-09-01
The pVACseq read split is hand-rolled where topiary provides split_reads_by_vaf


iskandr:
`pvacseq_metadata` computes the alt/ref split itself:

```python
alt = int(round(depth * vaf)) if vaf is not None else 0
...
counts=(n_total, n_alt_reads, max(0, n_total - n_alt_reads), n_alt_reads),
```

`vaxrank/external_input.py:1317` and `:1336`. topiary 5.37.0 exports `split_reads_by_vaf(depth, vaf)` returning `(n_alt_reads, n_ref_reads)`, which is the same arithmetic with the null handling already settled — both outputs are NA wherever either input is absent, rather than a zero standing in for an estimate nobody made.

The LENS path already reads topiary's derived columns rather than deriving (#374). This is the last hand-rolled copy.

## Why it is worth removing rather than leaving

It is a copy of an upstream rule living in this repo, which is the shape that produced the version bugs (#363, #367, #368, #369) and the read-count bug (#374). Each time, the local copy and the upstream rule agreed when written and diverged later — the local one being simpler is exactly why it drifts.

Two places they can already differ:

1. **Rounding.** `int(round(...))` is banker's rounding in Python 3; whatever topiary does may not match, and nothing checks.
2. **The reference count.** `max(0, n_total - n_alt_reads)` clamps a negative to zero, which hides an inconsistent depth/VAF pair rather than surfacing it. topiary returns NA when it cannot answer.

## Fix

Use `split_reads_by_vaf`. Its inputs are Series, so either build one per group or wrap the scalar call — worth checking which reads better against the existing loop rather than assuming.

Also fold in the fourth element of that tuple: `n_alt_reads_supporting_protein_sequence` is currently set to `n_alt_reads`, asserting that every variant-supporting read also spans the assembled protein sequence. On the LENS path that field holds a genuinely different count (`cds_overlap_reads`), so pVACseq is claiming a measurement it does not have.

iskandr:
Same rule as #379, #382 and #381: **a number must name its subject and its derivation, absent beats substituted, and the rule lives in one place.** This issue is the third clause — the arithmetic is duplicated rather than delegated.

[...]

Issue #382, opened 2026-09-01
n_alt_reads holds three different quantities depending on input path, and is labelled on only two


iskandr:
`MutantProteinFragment.n_alt_reads` feeds the combined-score DSL, where `sqrt(n_alt_reads) * target_epitope_score` is the documented canonical form. It currently holds a different quantity on each input path:

| path | what the field holds | `read_count_method` |
|---|---|---|
| native (isovar) | **fragments**, not reads — `num_alt_fragments` | `""` (#379) |
| LENS | depth × VAF, rounded — an estimate | `rna_depth_x_vaf` |
| pVACseq | depth × VAF, rounded — an estimate | `rna_depth_x_vaf` |

`sqrt()` weights all three identically.

## The fragment/read confusion is acknowledged in the code

```python
# TODO: distinguish reads and fragments in Vaxrank?
n_overlapping_reads=isovar_result.num_total_fragments,
n_alt_reads=isovar_result.num_alt_fragments,
n_ref_reads=isovar_result.num_ref_fragments,
n_alt_reads_supporting_protein_sequence=protein_sequence.num_supporting_fragments,
```

`vaxrank/mutant_protein_fragment.py:211`. Four fields named "reads", all populated with fragment counts. For paired-end data a fragment is two reads, so these are roughly half the read count for the same evidence — and the report prints them under "RNA reads supporting variant allele".

This was a fair question to defer when every run came through one path. It stopped being deferrable when #374 and #376 made the external paths carry read-based estimates into the same field, because now the field's meaning depends on where the run came from and nothing on the object says which.

## Consistent with the vocabulary elsewhere

#376 adopted topiary's convention that a derived number names its derivation (`rna_reads`, `rna_depth_x_vaf`, `cds_overlap_reads`, `tpm_x_dna_vaf`). Fragment counts are a fourth quantity with no term in that vocabulary, stored in a field whose name asserts a third thing.

Same problem shape as #381 and openvax/topiary#238 on the expression axis: the column name states a unit, not a subject, and two sources fill it with different subjects.

## Fix

1. Give the native path a `read_count_method` — either a new `rna_fragments` term (and propose it upstream so both packages share the vocabulary), or convert to reads at the boundary if that conversion is defensible.
2. Decide whether the fields should be renamed to say fragments where they hold fragments, or whether everything should be normalized to reads on the way in. Renaming is more honest; normalizing keeps `sqrt(n_alt_reads)` comparable across paths, which is what the ranking actually needs.
[...]

PR #239, opened 2026-04-23
Load real LENS schemas; emit per-predictor rows for DSL combination

iskandr:
## Summary

Makes `--input-lens` actually work on real LENS outputs (v1.4 / v1.5 / v1.9), surfaces every detected predictor to the Topiary DSL so formulas can combine them, and fails fast when a formula references a predictor / version / kind the data doesn't expose.

## What was broken

- **`load_lens()` expected columns no real LENS output produces** (`mhc_allele`, `binding_affinity`, `reference_peptide`, …). Against any actual LENS file you'd get "missing required columns" immediately.
- **Multiple predictors were invisible to formulas.** LENS v1.4/v1.5 emit netMHCpan + MHCflurry; v1.4 also emits netMHCstabpan (stability, a different topiary Kind). The old loader picked one and the others disappeared.
- **External-input scoring bypassed the DSL.** `write_neoepitope_report` called `EpitopePrediction.logistic_epitope_score()` directly, so `filter_expr` / `score_expr` were silently ignored for LENS and pVACseq. Formulas referencing missing fields produced `NaN` rather than errors.

### 1. Generic predictor sniffing

`load_lens` discovers predictors by regex on column names — `<tool>_<version>.<metric>` → bucket by `(tool, version)` → match against a small registry (`mhcflurry`, `netmhcpan`, `netmhcstabpan`) that declares each tool's topiary Kind (`pMHC_affinity` / `pMHC_stability`) and its value / percentile metric names. Unknown tools are ignored so future LENS versions don't hard-fail.

### 2. `--lens-predictor` gains `all`

- `auto` (default): pick a canonical pMHC_affinity predictor — `mhcflurry` preferred, falls back to `netmhcpan`.
- `all`: emit one `EpitopePrediction` per `(peptide, allele, detected predictor)`. Report DataFrame stays one row per `(peptide, allele)`; DSL sees all predictors.
- Any specific tool name: use only that, error if absent.

Allele strings are normalized from LENS's `HLA-A01:01` to `HLA-A*01:01`.

### 3. External-input scoring routes through Topiary DSL

`write_neoepitope_report` now builds a topiary long-format DataFrame (`predictions_to_topiary_df`), applies `build_filter_node` + `build_score_node`, and collapses per-group scores back onto the report rows. With no `filter_expr` / `score_expr` the default node reproduces the legacy `logistic_epitope_score` output byte-for-byte (regression tests cover both affinity and percentile_rank scoring modes).

This lets LENS users write combination formulas like:
[...]

PR #384, opened 2026-09-01
Say what n_alt_reads counted, on every path that fills it

### LINKED ISSUE #379: Native-pipeline fragments carry no read-count method or sequence source
#376 gave `MutantProteinFragment` a `read_count_method` and a `sequence_source`, and populates them on both external paths. The native VCF/BAM pipeline does not, so its fragments carry `""` for both.

That is the path most runs take, and it is the one where the distinction is sharpest: a native run's counts come from isovar against an alignment (`rna_reads` — genuinely counted), while a LENS run's come from depth × VAF (`rna_depth_x_vaf` — an estimate). The report now names the derivation for the estimate and says nothing for the measurement, which is backwards from what a reader would guess.

Same for `sequence_source`: `varcode_translation` and `isovar_assembly` are exactly the "translated from reference versus assembled from RNA" distinction, and the native path is where both actually occur.

## Was blocked, may not be

I recorded this in #376 as waiting on openvax/topiary#102. topiary 5.38.0 has since shipped `fragment_from_isovar_result` and `fragments_from_dataframe`, with `sequence_source` riding on the fragment — so the upstream half may be done. Worth checking whether vaxrank's native path can adopt those builders rather than constructing `MutantProteinFragment` itself, which would carry both fields across for free and remove a hand-written construction.

Not verified either way yet; filing so it is not lost, since #375 closed with this half explicitly out of scope.

iskandr:
Closes #379, #382, #383 — one problem in three places.

`n_alt_reads` feeds the combined-score DSL, where `sqrt(n_alt_reads) * target_epitope_score` is the documented canonical form. It held three different quantities and named none of them:

| path | quantity | label before | label after |
|---|---|---|---|
| native (isovar) | **fragments** | — | `rna_fragments` |
| LENS | depth × VAF | `rna_depth_x_vaf` | unchanged |
| pVACseq | depth × VAF, hand-rolled | `rna_depth_x_vaf` | topiary's `split_reads_by_vaf` |

### The native path is the one the field name misleads about

isovar counts fragments. For paired-end data a fragment is two reads, so those numbers are roughly **half** the read count for the same evidence, sitting in four fields named "reads". The code carried a `TODO` asking whether to distinguish them — fair to defer while one path existed, and not once #374/#376 put read-based estimates into the same field.

**Not converted:** the conversion needs to know whether the library was paired-end, and isovar does not say. Labelled instead, so a consumer sees the quantities differ rather than inferring it from a field name that asserts the wrong one.

`rna_fragments` is a **local extension**, marked as one, with a test that fails when topiary adopts a term (openvax/topiary#239) — same shape as the `is_named_version` stand-in that was deleted for the import once upstream landed.

### Sequence source on both native constructors

`isovar_assembly` and `varcode_translation`, read from `topiary.SEQUENCE_SOURCES` rather than restated, so a rename upstream fails at import instead of stamping an unrecognized term. The varcode path leaves `read_count_method` blank — no RNA was consulted, so its zeros are "nothing counted", not a derivation that produced zero.

### pVACseq's split is topiary's now

Two things the local copy got wrong beyond being a copy:

- `int(round(depth * vaf))` was never checked against topiary's rounding.
- The reference count was `max(0, depth - alt)`, which clamps a negative to zero — hiding an inconsistent depth/VAF pair instead of surfacing it.

Both halves come from `split_reads_by_vaf`, and alt + ref now sums to depth:

```
pvacseq  n_total 200  n_alt 68   n_ref 132  n_prot 0
lens     n_total 446  n_alt 4    n_ref 442  n_prot 9
```

`n_alt_reads_supporting_protein_sequence` was set to `n_alt_reads` on the pVACseq path, asserting every variant-supporting read also spans the assembled peptide — a measurement pVACseq never makes. Zero now; on the LENS path that field holds a genuinely different count (`cds_overlap_reads`).

### Verification

- 1168 tests pass, including 5 new ones in `tests/test_read_count_provenance.py`.

### DIFF
--- a/tests/test_read_count_provenance.py
+++ b/tests/test_read_count_provenance.py
@@ -127,3 +127,45 @@ def test_pvacseq_claims_no_protein_supporting_count():
     fragment = result.ranked[0][1][0].mutant_protein_fragment
 
     assert fragment.n_alt_reads_supporting_protein_sequence == 0
+
+
+def test_delegating_the_split_did_not_move_any_number():
+    """Migrating to topiary's arithmetic must be value-preserving.
+
+    The local copy was ``int(round(depth * vaf))``. Python's ``round`` is
+    banker's rounding, and nothing had checked that topiary agrees — so
+    this pins that it does, at the ``.5`` boundaries where the two could
+    differ. The reason to delegate is that a private copy of an upstream
+    rule drifts, not that this one had drifted.
+    """
+    import pandas as pd
+    from topiary import split_reads_by_vaf
+
+    cases = [(1, 0.5), (3, 0.5), (5, 0.5), (7, 0.5),
+             (2, 0.25), (10, 0.05), (100, 0.005)]
+    depth = pd.Series([c[0] for c in cases], dtype="Float64")
+    vaf = pd.Series([c[1] for c in cases], dtype="Float64")
+
+    alt, _ref = split_reads_by_vaf(depth, vaf)
+
+    assert [int(a) for a in alt] == [
+        int(round(d * v)) for d, v in cases]
+
+
+def test_an_absent_vaf_yields_no_split_rather_than_zero():
+    """Both halves are NA when either input is missing.
+
+    The local copy returned 0 for a missing VAF and derived the reference
+    count as ``max(0, depth - 0)`` — the whole depth, as if every read
+    supported the reference. topiary returns NA for both, which is what
+    lets the caller record "no estimate" instead of a confident zero.
+    """
+    import pandas as pd
+    from topiary import split_reads_by_vaf
+
+    alt, ref = split_reads_by_vaf(
+        pd.Series([100.0], dtype="Float64"),
+        pd.Series([None], dtype="Float64"))
+
+    assert pd.isna(alt.iloc[0])
+    assert pd.isna(ref.iloc[0])
--- a/tests/test_read_count_provenance.py
+++ b/tests/test_read_count_provenance.py
@@ -1,0 +1,129 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


iskandr:
Reviewed. One claim in the description was unverified and turns out to be wrong; correcting it.

### The rounding was fine

I listed `int(round(depth * vaf))` as something the local copy got wrong because it "was never checked against topiary's rounding". So I checked it. They agree — including at the `.5` boundaries where Python's banker's rounding is the interesting case:

```
depth=1   vaf=0.5    product=0.5   old=0   topiary=0
depth=3   vaf=0.5    product=1.5   old=2   topiary=2
depth=5   vaf=0.5    product=2.5   old=2   topiary=2
depth=7   vaf=0.5    product=3.5   old=4   topiary=4
```

And on every row of the pVACseq fixture: 3 rows compared, 0 differing.

**The migration is value-preserving.** The reason to delegate is that a private copy of an upstream rule drifts — not that this one had. The real defect the copy carried is the other one: `max(0, depth - alt)` clamping a negative to zero, and returning `0` for a missing VAF so the reference count became the whole depth, as if every read supported the reference.

Two tests added to pin both.

### Scoring impact, measured

The default combined score is `sqrt(n_alt_reads) * target_epitope_score` and does **not** name `n_alt_reads_supporting_protein_sequence`, so default runs are unaffected by the `n_prot = 0` change. `n_alt_reads` itself is unchanged on every fixture row.

Users who reference that field explicitly — the documented `log1p(n_alt_reads) * target_epitope_score + 0.1 * n_alt_reads_supporting_protein_sequence` — do see a change on pVACseq input:

```
SVVGSSSSS   old term 6.8   new 0.0
LATEKSRWS   old term 17.8  new 0.0
VVVGADGVG   old term 4.2   new 0.0
```

That is the correction, not a regression: the old value asserted that every variant-supporting read also spanned the assembled peptide, which pVACseq never measured. The term is now zero because the measurement is absent, which is what it should have been all along.

PR #391, opened 2026-09-02
Report other-allele reads only where the reference was counted

### LINKED ISSUE #389: "RNA reads supporting other alleles" is derived from a derived number, and is never a measurement on external input
`MutantProteinFragment.n_other_reads` is

```python
return self.n_overlapping_reads - (self.n_ref_reads + self.n_alt_reads)
```

and the report prints it as **"RNA reads supporting other alleles"**.

On both external paths `n_ref_reads` is not measured — it comes from topiary's `split_reads_by_vaf`, which computes `ref = depth - alt`. Substitute:

```
other = depth - (ref + alt) = depth - ((depth - alt) + alt) = 0
```

So the number is structurally zero whenever a VAF exists, and the report presents that zero as an observation about the locus. Measured on the real fixtures:

```
lens     (total, alt, ref, other)
         (446,   4,   442, 0)
         (2337,  51,  2286, 0)
         (67,    0,   0,   67)   <- no VAF
pvacseq  (200,   68,  132, 0)
         (350,   178, 172, 0)
```

The last LENS row is the worse case. No VAF means no split was made, so alt and ref are both 0 — and the subtraction reports **the entire depth as other-allele support**. A clean locus with 67 reads and no reported fraction is printed as 67 reads supporting a third allele, which reads as a messy locus and is the opposite of the truth.

## What the number should be

topiary answers this in 5.47.0 and the answer is to not have it: `n_rna_other` is **absent unless the source counted ref independently**. Neither of our external sources does, and the column is absent on both readers' frames accordingly.

The native isovar path is different — `num_ref_reads` is a real independent count there, so `n_other_reads` is a genuine measurement and should stay.

## Fix

Return `None` rather than a subtraction when the reference count was derived rather than counted, and omit the report row when it is `None`. `rna_evidence_method` already carries what is needed to tell: `rna_alignment` means counted, everything else means the ref side came from arithmetic.

Same rule as everywhere else in this series — absent beats substituted, and a derived value must not be presented as an observation.

iskandr:
Closes #389.

`n_other_reads` is `n_overlapping_reads - (n_ref_reads + n_alt_reads)`, printed as **"RNA reads supporting other alleles"**. On both external paths the reference count is not measured — topiary splits a reported depth and fraction as `ref = depth - alt`. Substitute and the subtraction collapses:

```
other = depth - (ref + alt) = depth - ((depth - alt) + alt) = 0
```

A structural zero, printed as an observation about the locus. Measured on the real fixtures before this change:

```
lens     (446,  4,   442,  0)
         (2337, 51,  2286, 0)
         (67,   0,   0,    67)   <- no VAF reported
pvacseq  (200,  68,  132,  0)
```

The last LENS row is the worse case: no fraction means no split was made, so alt and ref are both zero and the subtraction returns **the whole depth** — a clean locus rendered as one where every read supports a third allele.

It is `None` now unless the reference was counted, and the report omits the row. `rna_evidence_method` already distinguishes the two: `rna_alignment` means counted, everything else means the reference side came from arithmetic. The native isovar path counts reference reads independently and keeps the number.

topiary reached the same conclusion for its own frames in 5.47.0, where `n_rna_other` is absent unless the source counted the reference.

### 5.47.0 adoption, folded in

Not optional — the floor admits it and it renames what these paths read.

- **pVACseq reads topiary's `n_rna_*`** rather than deriving depth × VAF. The LENS path has done this since #374; pVACseq kept a local copy only because topiary published no derived counts for it, and 5.47.0 does. That was the last hand-rolled split (#383).
- **Source-prefixed columns**: the pVACseq depth and the LENS CDS-overlap count are read under both spellings.
- **LENS's unqualified fraction comes from `lens_vaf`** — deliberately *not* `rna_vaf`, which topiary also publishes on LENS frames as a verbatim copy of it. Reading that would inherit an assay assertion the file never made, which is exactly what `source_vaf` exists to avoid.

### Verification

- 1177 tests pass; ruff clean. Floor moves to `topiary>=5.47.0`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ mhctools>=3.13.3,<4.0.0
 # netmhcpan) without pre-subsetting the frame.
 # 5.16.0 adds read_pvacseq for both pVACseq all_epitopes flavors and
 # categorical DSL helpers used by pVACseq external-input scoring.
-topiary>=5.45.0,<6.0.0
+topiary>=5.47.0,<6.0.0
 roman
 jinja2>=3.1
 weasyprint>=62.0
--- a/tests/test_epitope_io.py
+++ b/tests/test_epitope_io.py
@@ -524,7 +524,7 @@ def test_pvacseq_source_keyed_filtered_scores_do_not_broadcast(tmp_path):
     report_df, epitopes = _loaded.report_df, list(_loaded.epitopes)
     csv_path = tmp_path / "filtered.csv"
     cfg = EpitopeConfig(
-        filter_expr="tumor_dna_vaf > 0.5",
+        filter_expr="pvacseq_tumor_dna_vaf > 0.5",
         score_expr="affinity.value")
 
     write_neoepitope_report(
--- a/tests/test_read_count_provenance.py
+++ b/tests/test_read_count_provenance.py
@@ -275,3 +275,58 @@ def test_pvacseq_keeps_its_qualified_dna_vaf():
 
     assert result.dna_vaf_by_variant
     assert result.source_vaf_by_variant == {}
+
+
+def test_other_allele_reads_are_absent_when_the_reference_was_derived():
+    """A subtraction of a subtraction is not an observation.
+
+    Where a source reports a depth and a fraction, the reference count is
+    ``depth - alt``, so ``depth - (ref + alt)`` is structurally zero — and
+    the report printed that zero as a fact about the locus. Worse when the
+    source reports no fraction: alt and ref are both 0 and the subtraction
+    returns the whole depth, rendering a clean locus as one where every
+    read supports a third allele.
+    """
+    config = EpitopeConfig()
+    report = read_pvacseq_report(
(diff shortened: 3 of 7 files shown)


iskandr:
Reviewed. One finding, fixed.

`reference_count_was_measured` hardcoded `rna_alignment`:

```python
return self.rna_evidence_method == RNA_EVIDENCE_METHOD_ALIGNMENT
```

That agrees with topiary today — only a direct count is `measured` — and it is a local copy of topiary's rule, so any measured method added upstream would be silently excluded. It asks `provenance_for_method` now, and a test walks every term in `READ_COUNT_METHODS` asserting the two agree rather than restating the expected answer.

Also checked:

- **Consumers**: only `report.py` (guarded) and a serialization round-trip test, which compares the property on both sides and is unaffected by `None`.
- **The no-RNA path**: varcode-only fragments have a blank method, so `reference_count_was_measured` is False and `n_other_reads` is `None`. That is right — no RNA was consulted, so the count is unknown rather than zero, which is what the old subtraction returned.

Issue #248, opened 2026-04-28
Per-modality length and antigen-count defaults via shared select_antigen_window helper


iskandr:
## Motivation

mRNA constructs concatenate antigens, peptide pools don't — so the natural per-antigen window length differs:

- **Peptide pools (PGV-001 style)**: 15-25 aa SLPs, one antigen per construct, ~20 constructs per vaccine.
- **mRNA constructs**: 15-20 aa per antigen window (shorter so concatenation fits in a single mRNA), 5 antigens per construct.

#246 added per-modality defaults and a candidates-per-slot field. This issue tracks **factoring the window-selection helper out of \`peptide.py\` so both modalities call the same code path**, and **ensuring config flows into ranking** so the ranking pipeline doesn't waste effort on windows the modality won't use.

## Current state (post-#246)

- \`vaxrank/peptide.py\` has \`_slp_window(fragment, base_name, max_length_aa)\` that mutation-centers a window.
- \`vaxrank/mrna.py\` calls \`peptide._slp_window\` cross-module — works but is a layering smell.
- Both modalities accept \`min_antigen_length_aa\` / \`max_antigen_length_aa\` / \`antigens_per_construct\` / \`max_constructs\` / \`candidates_per_slot\`.

## Refactor

1. **Extract** \`select_antigen_window(fragment, *, min_aa, max_aa)\` into \`vaxrank/vaccine_library.py\` (or a new \`vaccine_assembly.py\`).
2. Both \`peptide.py\` and \`mrna.py\` import it.
3. Remove the cross-module \`from .peptide import _slp_window\` in mrna.py.
4. Move shared modality-bookkeeping (the \`base_name\` formatter, the alt-suffix logic for \`candidates_per_slot\`) into the same shared module.

## Optional follow-up

Have the ranking pipeline (\`core_logic.py\`) honor a "target window length" hint so it doesn't have to enumerate every length 15..25 internally when only the modality's window matters. Today vaxrank uses \`vaccine_peptide_length\` as a single number; the modality re-windows downstream. Cleanest is to keep the ranking range wider (so different modalities can pick their own subset) and only narrow at output time — which is what #246 does. So this is an "if it ever matters" optimization.

## Related

- Follow-up to #245 / #246.

PR #392, opened 2026-09-02
Carry DNA evidence, mirroring RNA

iskandr:
Addresses the DNA half of openvax/topiary#248, which was written partly in answer to vaxrank asks.

vaxrank carried a DNA VAF and nothing else, so a report could show a fraction with no idea what depth produced it. **0.45 from two reads and 0.45 from two hundred are not the same claim** about how well a variant call is backed, and only the fraction reached the page.

topiary 5.47.0 derives DNA counts the same way it derives the RNA ones — from a reported depth and fraction — and labels them. `MutantProteinFragment` carries them now, read from the same winning row as the RNA counts:

```
pvacseq all_epitopes   dna(total, alt, ref) = (1233, 372, 861)
                       vaf 0.302   dna_depth_x_vaf
pvacseq aggregated     (None, None, None)   vaf 0.45   no method
lens                   (None, None, None)   vaf None   no method
```

### The two absences are the point

- The **aggregated** flavour reports a DNA fraction and no DNA depth, so deriving counts would invent the denominator.
- **LENS** states no assay for its single fraction at all, so giving it DNA evidence would assert the one thing that column cannot say — the same reason its RNA split is `rna_depth_x_source_vaf`.

The report gains DNA rows mirroring the RNA ones, each omitted rather than zero-filled where the source did not answer.

### Verification

- 1182 tests pass; ruff clean.

### DIFF
--- a/tests/test_read_count_provenance.py
+++ b/tests/test_read_count_provenance.py
@@ -465,3 +465,31 @@ def test_the_report_shows_dna_counts_only_where_they_exist():
     rows = variant_data_for(
         aggregated_peptide.mutant_protein_fragment, aggregated_peptide)
     assert 'DNA reads supporting variant allele' not in rows
+
+
+def test_the_reported_dna_vaf_comes_from_the_row_its_counts_came_from():
+    """One observation, not a fraction and a depth from different rows.
+
+    The report used to read the DNA VAF from a per-variant map while the
+    counts came from a specific row. They agree today, and nothing enforced
+    it — a variant group whose rows disagreed would print a fraction beside
+    a depth that never produced it, which is the per-field maximum problem
+    one field over.
+    """
+    from vaxrank.report import TemplateDataCreator
+
+    fragment = _all_epitopes_fragment()
+    creator = TemplateDataCreator.__new__(TemplateDataCreator)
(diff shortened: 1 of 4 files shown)


iskandr:
Reviewed. One finding, fixed.

**The DNA VAF and the DNA counts came from different places.** The report read the fraction from `dna_vaf_by_variant` — a per-variant map — while the counts this PR adds come from one specific winning row. They agree on every fixture:

```
all_epitopes   fragment.dna_vaf=0.302   map=0.302
aggregated     fragment.dna_vaf=0.45    map=0.45
```

and nothing enforced it. A variant group whose rows disagreed would print a fraction beside a depth that never produced it — the per-field maximum problem this file already guards against for the RNA counts, reintroduced one field over.

The fragment's own value wins now, with the map as fallback for sources that report a fraction and no counts. A test feeds a deliberately contradictory map and asserts the fragment's value reaches the page.

Also checked:

- **Serialization**: the six new fields are dataclass fields, so they round-trip. Verified `n_dna_alt` and `dna_evidence_method` survive `to_dict`/`from_dict`.
- **No fabrication on either absence path**: aggregated pVACseq keeps its fraction and gets no counts; LENS gets neither. Both pinned.

PR #393, opened 2026-09-02
Give the combined score a threshold that means one thing everywhere

### LINKED ISSUE #385: The combined-score namespace exposes n_alt_reads without its subject
`MutantProteinFragment` now records `read_count_subject` — `fragments` on the isovar path, `reads` on both external paths (#385). The combined-score DSL namespace does not, and it is where the harm actually lands.

```python
'n_alt_reads': float(frag.n_alt_reads or 0),
```

`vaxrank/combined_score_dsl.py`. So a config saying

```
sqrt(n_alt_reads) * target_epitope_score
```

means "square root of a fragment count" on a native run and "square root of a read estimate" on a LENS run, and any documented threshold — `n_alt_reads > 5` — is a different bar depending on where the file came from. Within one run the unit is consistent and the ranking does not change; the harm is entirely in what travels between runs.

## Why this was not fixed alongside the field

Making `n_alt_reads` unavailable on the isovar path would express the truth and break every existing config. `count_in` returns `None` for the wrong subject, which is right for an accessor and wrong for an arithmetic namespace: `sqrt(None)` raises, and coercing to `0` reintroduces exactly the substitution the subject axis exists to prevent — an unanswerable count becoming "no evidence" rather than "different unit".

## Options

1. **Add subject-qualified names** alongside: expose `n_alt_fragments` and `n_alt_reads`, each populated only when the subject matches, and make referencing the absent one a loud eval error rather than a silent zero. Existing configs keep working; a portable config can be written for the first time.
2. **Expose the subject as a namespace value** (`read_count_subject`) so an expression can branch. Honest, but pushes the problem onto every config author.
3. **Leave it and document.** The status quo, which is a footgun that only fires across runs — the hardest kind to notice.

Leaning 1. It is additive, it makes the wrong reference fail rather than mislead, and it matches how the report now names both axes.

Whichever, the deprecation question needs answering: does bare `n_alt_reads` eventually warn, and against what — being ambiguous, or being subject-free?

iskandr:
Closes #385.

`sqrt(n_alt_reads) * target_epitope_score` is the documented canonical form, and `n_alt_reads` holds a **different quantity per input path**: isovar fragments on the native path, read estimates on both external ones. `sqrt()` weights them identically, so a documented threshold or a config copied between projects silently changes what it measures.

`n_rna_alt` / `n_rna_ref` / `n_rna_overlapping` / `n_rna_supporting_protein_sequence` are bound now — each is whatever the source counted, so a threshold against one means the same thing everywhere. On an isovar fragment the difference is visible:

```
n_alt_reads  15    reads, isovar's own
n_rna_alt     8    fragments, what it counted
```

### The unit is deliberately not bound

I added `rna_evidence_subject` as a binding first, then removed it: this grammar is **arithmetic only**, so a string is a name no expression can use, and the validator rejects the comparison that would consume it —

```
ValueError: Disallowed expression element Compare ...
```

Binding it would advertise a capability the grammar forbids. `rna_evidence_subject` is on the fragment and in the report, which is where a config author reads which bar their threshold cleared.

### Also

- `n_dna_alt` / `n_dna_ref` / `n_dna_overlapping` follow #392, so an expression can weight how well a *call* is backed and not only how well it is expressed.
- The `*_reads` names stay — existing configs name them — and their docstrings now say what they actually hold rather than what their names suggest.
- Counts are read with `getattr`: callers pass stand-ins for the fragment as well as the real dataclass, and a missing name binds `0` the way a real fragment with no RNA does.

### Verification

- 1185 tests pass; ruff clean.

### DIFF
--- a/tests/test_combined_score_dsl.py
+++ b/tests/test_combined_score_dsl.py
@@ -315,3 +315,73 @@ def test_dsl_eval_error_warns_when_headline_binding_missing():
                 "binding; got: %r" % [r.getMessage() for r in warnings])
     finally:
         dsl_mod.combined_score_bindings = real_extractor
+
+
+def _isovar_backed_peptide():
+    """A stub vaccine peptide whose fragment came from isovar.
+
+    Only the attributes ``combined_score_bindings`` reads are needed, and
+    a real VaccinePeptide would drag in epitope selection irrelevant here.
+    """
+    import types
+
+    from vaxrank.mutant_protein_fragment import MutantProteinFragment
+
+    protein_sequence = types.SimpleNamespace(
+        gene_name="BRAF", amino_acids="SIINFEKLAA",
+        mutation_start_idx=2, mutation_end_idx=3,
+        num_supporting_fragments=7, num_supporting_reads=13, transcripts=[])
+    fragment = MutantProteinFragment.from_isovar_result(types.SimpleNamespace(
+        variant=object(), top_protein_sequence=protein_sequence,
+        num_total_fragments=20, num_alt_fragments=8, num_ref_fragments=12,
+        num_total_reads=38, num_alt_reads=15, num_ref_reads=23))
+    return types.SimpleNamespace(
+        target_epitope_score=1.0, self_epitope_score=0.0,
+        expression_score=1.0, mutant_protein_fragment=fragment)
+
+
+def test_the_portable_name_reports_what_the_source_counted():
+    """``n_rna_alt`` is the unit the run actually has; ``n_alt_reads`` is not.
+
+    On the isovar path the two are different numbers for the same
+    evidence — 8 fragments and 15 reads — and ``sqrt()`` weights whichever
+    it is given identically. A threshold naming ``n_alt_reads`` therefore
+    means a different bar depending on where the file came from, which is
+    the portability problem (#385). ``n_rna_alt`` means one thing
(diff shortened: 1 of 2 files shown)


iskandr:
Reviewed. No finding that needs a code change; two things checked and one question raised.

**No collisions, default untouched.** 15 bindings, no duplicate names, and `n_alt_reads` still binds exactly what it did:

```
default expr:   sqrt(n_alt_reads) * target_epitope_score
names it uses:  ['n_alt_reads', 'target_epitope_score']
n_alt_reads unchanged: True
```

**The string bindings were a genuine mistake, caught before merge.** I bound `rna_evidence_subject` so an expression could branch on the unit, then found the validator rejects the comparison that would consume it — this grammar is arithmetic only. Binding it would have advertised a capability the grammar forbids. Removed, with the reasoning in the docstring so the next person doesn't re-add it.

**Question this PR raises but should not decide:** the default still names `n_alt_reads`, the non-portable half — so the documented canonical score is `sqrt(fragments)` on the native path and `sqrt(reads)` on the external ones. Filed as #394 rather than changed here, for the same reason `allele_free_evidence` shipped defaulting to `all`: ship the mechanism, decide the semantics with evidence. There is no isovar fixture in this repo to measure the reordering against, which is worth its own attention.

PR #395, opened 2026-09-02
Reject an unavailable column when the config is read

### LINKED ISSUE #388: A DSL expression naming an unavailable annotation column fails at eval, not at config time
Since #380, LENS annotation columns are addressable in filter and score expressions. A config naming one the frame does not have fails, loudly and with a good message — but from topiary at eval time, not from `validate_dsl_against_predictions`:

```
ValueError: Column 'no_such_column' not found in DataFrame.
Available columns: ['affinity', 'allele', 'antigen_source', 'c_flank', 'ccf', ...]
```

That is not a silent failure and is fine as a floor. The gap is the same one `validate_default_methods` already exists to close: topiary can report what the data lacks, which is the half it can know; *which config key to edit* is the half it cannot, and is the half the user needs.

## Why it is more than tidiness now

The available columns depend on the source, and topiary 5.46.0 makes the rule explicit: the evidence columns are the same **vocabulary**, not the same **columns**. Verified across both pVACseq flavours:

```
pvacseq aggregated    missing: ['gene_expression']
pvacseq all_epitopes  missing: none
```

So `gene_expression > 1` is a working config on one pVACseq export and a crash on another, from the same tool, and the failure arrives mid-run rather than when the config is read. LENS files vary the same way — a file with no `tpm` column produces no `gene_expression` either.

## Fix

Extend `validate_dsl_against_predictions` to check bare column references against the frame, using `topiary.available_evidence_columns(df)` for the evidence vocabulary plus the frame's own columns for passthrough annotations. Raise naming the config key and the expression, the way the kind/method/version checks already do.

Worth keeping topiary's "did you mean" list in the message — it is better than anything vaxrank would generate, and the point is to add the config-key half rather than to replace what is already good.

iskandr:
Closes #388.

Since #380, LENS annotation columns are addressable in filter and score expressions. A config naming one the input lacks failed loudly — but from topiary, at eval time, and without saying which setting produced the reference.

That last part is the half topiary cannot know. It reports what the data lacks, with a better "did you mean" list than vaxrank would generate; **which config key to edit** is what the user needs, and is the same division `validate_default_methods` already makes.

### The timing mattered more than it looked

Validation ran only in the report writer, and both readers score during load — so a user met the error mid-run rather than when their config was read. It runs before scoring now, on every path.

```
epitopes.score_expr references column 'nope', which the loaded input does not
carry. Expression: '...'. Columns available on this input: [...]. Which columns
exist depends on the source and its flavour, so an expression that works on one
export may not on another.
```

That closing sentence is not decoration. Which columns exist genuinely varies: pVACseq's **aggregated** flavour carries no `gene_expression` where **all_epitopes** does, so the same expression is a working config on one export from one tool and a crash on another.

`collect_dsl_references` already returned the column set and nothing read it — its docstring said topiary validated these, which was true and not sufficient.

### Verification

- 1188 tests pass; ruff clean.

### DIFF
--- a/tests/test_epitope_dsl.py
+++ b/tests/test_epitope_dsl.py
@@ -1057,3 +1057,75 @@ def test_a_caller_supplied_frame_cannot_add_a_phantom_allele(unstated):
 
     assert dict(scored.per_allele_scores) == {
         "HLA-A*02:01": pytest.approx(0.8)}
+
+
+def test_an_unavailable_column_fails_when_the_config_is_read(tmp_path):
+    """Naming a column the input lacks must fail early and name the setting.
+
+    topiary raises on this too, with a better "did you mean" list than
+    vaxrank would generate — but at eval time, mid-run, and without saying
+    which config key produced the reference. That is the half topiary
+    cannot know and the half the user needs (#388).
+    """
+    from vaxrank.epitope_config import EpitopeConfig
+    from vaxrank.epitope_io import read_lens_report
+
+    path = tmp_path / "annotated.tsv"
+    path.write_text(
+        "peptide\tallele\tpep_context\ttpm\tmhcflurry_2.1.1.aff\n"
+        "SIINFEKL\tHLA-A02:01\tXXSIINFEKLXX\t50\t20\n")
+
+    config = EpitopeConfig(
+        score_expr=(
+            "affinity.value.logistic_normalized(350,150) * (nope > 1)"))
+
+    with pytest.raises(ValueError) as excinfo:
+        read_lens_report(path, epitope_config=config)
+
+    message = str(excinfo.value)
+    assert "epitopes.score_expr" in message      # which key to edit
+    assert "nope" in message                     # what was wrong
+    assert "gene_tpm" in message                 # what is available
+
+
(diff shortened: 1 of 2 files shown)


iskandr:
Reviewed. No code change needed; three things checked.

**It does not over-reject.** Every column the frame genuinely carries is still accepted, including the ones added by recent topiary work:

```
accepted: affinity.value * (n_rna_alt > 1)
accepted: affinity.value * (peptide_length > 8)
accepted: affinity.value * (gene_tpm > 1)
```

**It deliberately does not type-check.** `rna_evidence_subject > 1` passes this check — the column exists — and then fails at eval with topiary's error:

```
TypeError: Column 'rna_evidence_subject' contains non-numeric values
(could not convert string to float: 'reads'). Only numeric columns can be
used in DSL expressions.
```

That is the right division. Which columns are numeric is topiary's rule, and its message already names the column and the offending value; adding a numeric check here would be a local copy of that rule, which is the shape this whole series has been removing. The gap this PR closes is only the one topiary cannot fill: naming the setting.

**Cost is negligible.** Validation now runs on every load rather than only in the report writer; a full LENS fixture load is 0.06s, and the work is parsing two expressions that were parsed anyway.

PR #396, opened 2026-09-02
Read the LENS read-count column under the name topiary now uses

### LINKED ISSUE #390: topiary 5.47/5.48 renamed reader columns vaxrank reads by name (2 fail silently)
vaxrank pins `topiary>=5.45.0,<6.0.0`, so `pip install -U` picks up 5.47.0 /
5.48.0, which renamed reader columns vaxrank reads by name. Filed from a
topiary session — not touching this repo's code.

topiary now exports the mapping, so each fix is a lookup rather than a guess:

```python
from topiary import RENAMED_COLUMNS, renamed_column
renamed_column("vaf")  # 'lens_vaf'
```

## Sites, worst first

**1. `vaxrank/external_input.py:1106` — silent, and misattributes the cause**

```python
if cells.missing(r.get('rna_reads_covering_genomic_origin'))
```

Now `lens_rna_reads_covering_genomic_origin`. Every SNV/INDEL row counts as
lacking RNA reads, and line 1126 logs *"Likely upstream LENS bug"* — blaming
LENS for a topiary rename.

**2. `vaxrank/external_prediction.py:41` — silent zero**

```python
_PVACSEQ_RNA_DEPTH_COLUMNS = ("rna_depth", "tumor_rna_depth")
```

Now `pvacseq_tumor_rna_depth`. Neither name matches, so
`external_input.py:1232` falls to `default=0` and RNA depth becomes 0 with no
error — it just changes rankings.

Note `rna_depth` is not a topiary column and never was; topiary's canonical
name is `n_rna_overlapping` ("depth" presumes reads, and the count may be
fragments). Worth adding to the tuple.

**3. `vaxrank/external_input.py:745` — silent drop**

```python
value = row.get("vaf")
```

Now `lens_vaf`. `cells.missing` is true, the loop `continue`s, and the VAF
vanishes from the report. **Do not reach for `rna_vaf` here** — that is the
new canonical cross-source fraction; `lens_vaf` is LENS's own, assay
unstated, which is what the surrounding comment is careful about.

**4. `vaxrank/external_input.py:610`**

`row.get('rna_reads_covering_genomic_origin_with_peptide_cds')` →
`lens_rna_reads_covering_genomic_origin_with_peptide_cds`.

**5. `vaxrank/external_prediction.py:42-43` — already fine, worth simplifying**

```python
_PVACSEQ_RNA_VAF_COLUMNS = ("rna_vaf", "tumor_rna_vaf")
_PVACSEQ_DNA_VAF_COLUMNS = ("dna_vaf", "tumor_dna_vaf")
```

The first choice in each now exists: 5.47.0 added canonical `rna_vaf` /
`dna_vaf` derived under a stated method. The stale second entries are dead.

**6. `tests/test_epitope_io.py:527`**

`filter_expr="tumor_dna_vaf > 0.5"` raises. The message names the
replacement.

## Not affected

`MutantProteinFragment.n_alt_reads` and siblings are vaxrank's own fields fed
from `isovar_result`, not topiary columns — topiary renamed *its* copies to
`n_rna_alt_reads` etc., which does not reach vaxrank. Worth knowing the two
now differ in name for the same quantity.

`attach_read_evidence` → `attach_rna_evidence` and
`topiary.rna_evidence` → `topiary.evidence`: no vaxrank references.

## Also new and possibly useful

- `n_dna_alt` / `n_dna_ref` / `n_dna_other` / `n_dna_overlapping` / `dna_vaf`,
  mirroring the RNA set, from pVACseq's DNA depth and VAF.
- `n_rna_other` / `n_dna_other`: support for neither allele.
- A column is now **omitted rather than nulled** where a source cannot answer.
  `gene_expression > 1` on a frame where nobody supplied expression now
  raises instead of evaluating to NaN and silently emptying the frame.

iskandr:
Closes #390.

topiary 5.47.0 prefixes each tool's own columns with the tool name. vaxrank's SNV/INDEL completeness check read only the bare spelling, so **every such row counted as lacking RNA reads**:

```
SNV/INDEL rows: 10   counted as lacking RNA reads: 10
under the name topiary now uses:                    0
```

The data was fully present. The warning sent the reader upstream to look for a defect in the LENS file that was not there — worse than a silent miscount, because it spends someone else's time on a rename.

Two of #390's three sites were already fixed while adopting 5.47.0 in #391 (the pVACseq depth tuple and the LENS `vaf` read); this is the third.

The pVACseq depth lookup also gains `n_rna_overlapping` at the front — topiary's canonical name for that count. "Depth" presumes reads and the number may be fragments, which is why topiary does not call it depth.

### Verification

- 1190 tests pass; ruff clean.

### DIFF
--- a/tests/test_external_input.py
+++ b/tests/test_external_input.py
@@ -2018,3 +2018,45 @@ def test_a_pvacseq_row_with_no_rna_vaf_claims_no_derivation():
     assert pvacseq_rna_vaf({"rna_vaf": 0.0}) == 0.0
     assert pvacseq_rna_vaf({}) is None
     assert pvacseq_rna_vaf({"rna_vaf": ""}) is None
+
+
+def test_present_rna_read_counts_are_not_reported_as_missing():
+    """A renamed column must not be reported as absent data.
+
+    topiary 5.47.0 prefixes each tool's own columns with the tool name.
+    Reading only the bare spelling counted every SNV/INDEL row as lacking
+    RNA reads and warned about it — a rename surfaced as a defect in the
+    input file, which is the worst kind of wrong: it sends the reader
+    upstream to look for a bug that is not there (#390).
+    """
+    from topiary import read_lens
+
+    from vaxrank import cells
+
+    frame = read_lens(os.path.join(
+        DATA_DIR, "real_lens_subsets", "lens_v1.9_real_subset.tsv")).df
+    rows = [
+        row for row in frame.to_dict("records")
+        if str(row.get("antigen_source", "")).upper() in {"SNV", "INDEL"}]
+    assert rows, "fixture must contain SNV/INDEL rows for this to mean anything"
+
+    missing = [
+        row for row in rows
+        if cells.missing(cells.first(
+            row,
+            "lens_rna_reads_covering_genomic_origin",
+            "rna_reads_covering_genomic_origin"))]
+
+    assert missing == []
+
+
+def test_the_pvacseq_depth_lookup_prefers_the_canonical_name():
+    """``n_rna_overlapping`` first, because "depth" presumes reads.
+
+    topiary's count may be fragments, which is why its own name for this
+    is not "depth". Reading the canonical column first means the lookup
+    keeps working when a source's own spelling changes again.
+    """
+    from vaxrank.external_prediction import _PVACSEQ_RNA_DEPTH_COLUMNS
+
+    assert _PVACSEQ_RNA_DEPTH_COLUMNS[0] == "n_rna_overlapping"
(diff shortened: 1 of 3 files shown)


iskandr:
Reviewed. No further finding; one sweep worth recording.

**Checked for other columns with the same exposure.** Every name vaxrank reads by string literal from a LENS row, cross-referenced against what topiary 5.47.0 actually emits, looking for names that now exist *only* under a `lens_` prefix:

```
column names vaxrank reads that exist ONLY as lens_*: ['vaf']
```

and that one is already the fallback half of a pair fixed in #391:

```python
value = cells.first(row, "lens_vaf", "vaf")
```

So the three sites in #390 are the complete set — not three examples of a larger problem.

**On the warning text.** #390 describes this site as logging *"Likely upstream LENS bug"*. That phrase is on the adjacent `gene_name` and `transcript_id` warnings rather than this one, which said *"those kinds are expected to carry RNA-read counts"*. The misdiagnosis is real either way — both send a reader upstream — and the reworded message now names the column it actually looked for, so the next rename is diagnosable from the log line rather than from the source.

Issue #414, opened 2026-09-07
Test osteosarc vaccine peptide selection against documented sequences after Isovar integration


iskandr:
Add offline Vaxrank regression tests comparing its selected vaccine peptides with independently documented selected sequences from osteosarc.com, after the relevant Isovar data and protein-validation changes have merged.

## Upstream dependency

- openvax/isovar#216 is merged: original RNA fixtures include six osteosarcoma loci across bulk and nanopore tumor RNA.
- openvax/isovar#219 is still open: the active follow-up adds protein-level validation and addresses the spliced-indel failures tracked in openvax/isovar#215 and openvax/isovar#217. Begin downstream adoption from the reviewed merged commit and its published release, not the current dirty development tree.
- openvax/isovar#218 tracks expansion to every vaccine-included variant across all available RNA sources. Start with fully documented merged cases and extend coverage as that corpus lands; completing the entire expansion need not block an independently complete pilot regression.

## Required comparison

1. Preserve small original-read fixtures and offline reference assets from a pinned Isovar commit, with checksums, source provenance, assembly, annotation release, transcript versions, RNA sample/timepoint and variant alleles.
2. Independently snapshot the exact known selected amino-acid sequences, vaccine/provider/version, and original source URLs. Vaccine membership and a protein-change label alone are insufficient expectations. The DYNC1H1 page, for example, publishes different long-peptide selections for JLF, CeGaT and mRNA plus a minimal epitope: https://osteosarc.com/variant/DYNC1H1-chr14-101980529/. Preserve construct-added residues and peptide boundaries explicitly; do not silently treat a modified construct as an RNA-derived window.
3. Exercise real Isovar reconstruction through Vaxrank's fragment conversion, peptide-window generation, filtering, ranking and final selected VaccinePeptide sequences. Assert exact selected amino-acid sequences and mutation positions/junctions for reproducible cases, along with variant identity and supporting evidence. A nonempty candidate list or expected mutation somewhere in a peptide does not establish selection correctness.
4. Pin patient HLA alleles, peptide lengths, ranking/selection settings, relevant transcript/haplotype context and predictor versions. Use fixed, provenance-bearing predictions through the existing Topiary cache interface for deterministic offline tests; do not fabricate prediction scores to force agreement, regenerate the expected sequences with Vaxrank itself, or allow network/model downloads in CI. Include enough competing candidates to exercise selection rather than trivially selecting the only supplied peptide.
[...]

PR #431, opened 2026-09-09
Support offline references with content-keyed membership caches

### LINKED ISSUE #430: Offline Genome references crash during final peptide selection and need distinct cache identities
The new #414 original-read selection integration reaches `vaccine_peptides_for_variant -> predict_epitopes -> ReferenceProteome -> load_kmer_set_index` and crashes with `AttributeError: 'Genome' object has no attribute 'species'` at reference_proteome.py:369. Plain pyensembl.Genome is the supported type for an explicit offline GTF/cDNA/protein reference; it has annotation_version, not EnsemblRelease's species/release interface. Existing reconstruction-only Sid tests never reached MHC/self-reference selection.

The disk filename and build logging make the same assumption. Fix generic-reference support without assigning fabricated species/release attributes or colliding subset and full-reference caches. Reuse the existing content-based dataset identity, or actual protein-content identity when needed. Add a real offline-Genome end-to-end regression and distinct-reference cache contamination coverage. Related #414.

iskandr:
## Summary
Closes #430. Found while implementing #414; this prerequisite is intentionally separate from Topiary #296, which currently blocks cached final-selection tests.

- Support real offline pyensembl.Genome references without inventing species/release attributes.
- Key both memory and disk membership caches by existing content-based dataset identity (or actual proteins for custom providers), so distinct subsets cannot silently share results.
- Never reuse old species/release-only files; preserve them on disk and document the one-time rebuild.
- Include a real Sid-reference regression, different-content/same-label memory+disk regressions, legacy-cache rejection, and bump to 3.15.1.

## Verification
- Lint and diff checks passed.
- Focused reference suite: 57 passed against published Isovar 1.8.1, Topiary 5.53.0, mhctools 3.39.0 and Varcode 7.0.0.
- Full isolated suite and CI running. Will review their results, smoke the CLI, then merge and deploy only after all gates pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -920,6 +920,13 @@ Run linting and tests:
 
 The first run of the tests may take a while to build the reference proteome kmer index, but subsequent runs will use the cached index.
 
+Reference membership caches are keyed by the installed annotation/sequence
+content (or actual protein content for custom reference providers), not only
+species and release. This supports explicit offline PyEnsembl `Genome`
+references and prevents a subset from sharing a full-reference cache. Upgrading
+from species/release-only caches triggers a one-time rebuild; old cache files
+are not trusted or automatically deleted.
+
 ### Scripts
 
 - `develop.sh`: installs the package in editable mode and sets `PYTHONPATH` to the repo root.
--- a/tests/test_reference_proteome.py
+++ b/tests/test_reference_proteome.py
@@ -269,7 +269,7 @@ def test_load_kmer_set_builds_index_when_not_cached():
             ok_("ABCDEFGH" in kmers)
 
             # Cache file should now exist (compressed)
-            cache_path = os.path.join(tmpdir, "test_species_100_kmer_set_8_8.pkl.gz")
+            cache_path = kmer_set_index_path(genome, 8, 8)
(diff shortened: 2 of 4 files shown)


iskandr:
Final review complete. Lint and diff checks passed; the isolated final-commit full suite passed all 1,336 tests against published dependencies; the B16 smoke produced nonempty PDF, HTML, ASCII, JSON, CSV and both Excel reports. All Python 3.10–3.13 branch and PR CI jobs are green. This PR contains only the cache identity/Genome fix, its tests, documentation and version bump; #414's pending Topiary-cache tests are not included. Proceeding with merge and immediate deployment.

iskandr:
Vaxrank 3.15.1 is deployed from merge commit b69c506a662f5fde2533a40213f70f335a7fe3f7. deploy.sh passed lint and all 1,336 tests on clean main, built/uploaded both artifacts and pushed v3.15.1. pip index confirms 3.15.1; wheel and sdist SHA-256 values independently match local builds byte-for-byte. main/origin/main/tag agree and the release worktree is clean. #430 is closed. Post-merge test CI is running; the PR and branch matrices were fully green before merge.

PR #432, opened 2026-09-09
Validate malformed allele evidence at load time, not report time

### LINKED ISSUE #355: Report generation aborts on a malformed allele where it used to render blank
#345 turned two display-level unknowns into exceptions that abort report generation.

- `report.epitope_report_row_inputs` raises `"{kind} report evidence requires a patient allele"` for an allele-scoped leaf with a blank allele.
- `report.epitope_data` raises `"Template epitope rows require an explicit patient allele"` for the same.

Both were previously rendered as a blank cell. Both are correct as *invariants* — an allele-scoped prediction without an allele is malformed. The objection is where they fire: at render time, one such leaf anywhere in a run kills the entire PDF/HTML/ASCII report, and the operator is told nothing about which row or file produced it.

Enforce at load, where the fix is actionable and the offending row can be named, and let the renderer show a blank cell. `vaxrank.allele_evidence.is_allele_scoped` (added in #348) is now the single definition of which kinds must carry an allele, so the load-time check has something to test against.

Reachability is low — mhctools and both external readers always supply an allele — so this is a robustness fix, not an active bug.

Smaller, same area: `epitope_data`'s `Score` column now mixes floats with `'—'` where it was previously always numeric, which changes that column's dtype in the ASCII and XLSX tables.

iskandr:
## Summary
Closes #355. Bumps Vaxrank to 3.15.2.

- Centralize allele validation with Topiary kind scope and mhcgnomes, including non-human alleles and class-II pairs.
- Reject malformed scoped leaves at native CSV/JSON, LENS, pVACseq, prior-report reload, and source/WT predictor boundaries with available file/row/comparator provenance.
- Retain genuinely allele-free processing evidence without inventing patient typing.
- Keep directly supplied legacy report objects renderable, with blank unknown allele/score cells and numeric missing scores. Format only at the HTML/ASCII presentation boundary.

## Review and verification
- Reviewed validation placement, nested non-WT comparators, predictor exception boundaries, and unchanged scoring policy.
- 61 dedicated regression cases, including actual external loaders and packaged HTML/ASCII report assembly.
- Lint and git diff --check pass.
- Full isolated suite: 1,397 passed (29 warnings).
- B16 end-to-end smoke passed: nonempty PDF, HTML, ASCII, XLSX, neoepitope XLSX, JSON, and CSV outputs.
- CI must pass before merge; deployment and published-artifact verification follow immediately after merge.

### DIFF
--- a/ALLELE_VALIDATION.md
+++ b/ALLELE_VALIDATION.md
@@ -1,0 +1,16 @@
+# #355: validate allele evidence at input boundaries
+
+- Use the existing Topiary-backed `is_allele_scoped_kind` definition; do not
+  mistake allele-free processing or half-life observations for malformed MHC
+  evidence. Parse stated allele identifiers with mhcgnomes, including non-human
+  alleles and class-II pairs, without hand-written HLA prefix rules.
+- Reject malformed allele-scoped leaves before filtering/scoring in native
+  flat and serialized inputs, LENS and pVACseq. Include the input filename,
+  row and offending kind/comparator where those are available.
+- Keep rendering defensive for directly constructed legacy objects: show an
+  unavailable allele/score, never invent an allele, a zero or patient evidence.
+  Missing scores remain numeric missing values in tabular output.
+- Test nested non-WT comparators, blank/null spellings, valid class-II and mouse
+  alleles, allele-free kinds, report rendering and actual load paths. No ranking
+  policy changes. Review, lint, full tests, B16 smoke, green CI and deployment
+  are required before closing the issue.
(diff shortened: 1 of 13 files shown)


iskandr:
Shipped as Vaxrank 3.15.2. All eight Python 3.10–3.13 branch/PR test jobs passed before merge. Clean-main deployment reran lint and all 1,397 tests, built and checked wheel/sdist, uploaded both, and pushed v3.15.2. Independent PyPI JSON SHA-256 comparisons match both local artifacts. main, origin/main and v3.15.2 all point to 7c48bf6881978978b91815e99b804edf3d010cd3; release worktree is clean. B16 smoke produced nonempty PDF, HTML, ASCII, XLSX, neoepitope XLSX, JSON and CSV reports. Post-merge test CI is running: https://github.com/openvax/vaxrank/actions/runs/34419517532.

Issue #453, opened 2026-09-15
vaxrank tests fail against current mhctools releases (MeasurementContext, then unit): 8 failures on 3.44.10, 38 on 3.44.18


iskandr:
Eight parametrizations of `test_every_allele_free_kind_reaches_every_patient_allele` (`tests/test_epitope_dsl.py:760-775`) fail on `main` (84f9ebc) against mhctools 3.44.10, which is the current PyPI release and satisfies vaxrank's own `mhctools>=3.44.0,<4.0.0`.

```
$ python -m pytest tests/test_epitope_dsl.py -k every_allele_free_kind -n 0 -q
8 failed, 7 passed, 83 deselected

tests/test_epitope_dsl.py:771: in test_every_allele_free_kind_reaches_every_patient_allele
mhctools/pred.py:395: in __post_init__
E   ValueError: plasma_half_life predictions require MeasurementContext
```

Failing kinds: `cellular_uptake`, `cpp_classification`, `distribution_volume`, `plasma_half_life`, `systemic_clearance`, `systemic_elimination_half_life`, `systemic_exposure`, `tissue_concentration`.

The test builds an allele-free `Prediction` with no context:

```python
predictions=(Prediction(
    kind=kind, score=0.8, peptide="SIINFEKL", allele="",
    predictor_name="mhcflurry", predictor_version="2.1.1"),)
```

mhctools now requires a `MeasurementContext` for these kinds (`mhctools/pred.py:395`, added in 4c39b6b "Define peptide exposure result semantics", released in v3.44.10).

Not caused by varcode: the same 8 tests fail identically with varcode 7.3.0 and 8.0.0, so this is a vaxrank/mhctools mismatch.

Either the test should construct the `MeasurementContext` these kinds now need, or `attach_per_allele_scores` needs to handle context-free exposure predictions and the pin should move to the mhctools version that introduced the requirement.

iskandr:
This has grown well past the 8 `test_epitope_dsl` failures in the title. CI installs the newest mhctools, and 3.44.11–3.44.19 all shipped in the last few hours; against **mhctools 3.44.18** the suite fails **38** tests across 7 files, with a different message.

Measured on a clean `main` (84f9ebc), mhctools 3.44.18 shadowed onto an otherwise-current stack:

```
38 failed, 1527 passed in 70s

  14  tests/test_peptide_context.py
  10  tests/test_epitope_dsl.py
   6  tests/test_osteosarc_selection.py
[...]

PR #437, opened 2026-09-10
Audit Sid constructs with full RNA, cleavage and self-source evidence

iskandr:
## Summary

Closes #423 with the independently documented DYNC1H1 pilot and explicit unassessed coverage. Closes #436. Closes #438. Closes #439. These fixes were found during the real full-proteome audit and CI review. #414 remains open: reconstruction/construct processing is not historical final-selection agreement.

- Pin four published DYNC1H1 vaccine entries, their independently intended `KRFHATISF` ligand, clinical HLA typing and unresolved chemistry/provider facts. JLF V2/V3's terminal `KK` remains separate from native/RNA coordinates; solubility rationale is user-reported and CSBio's role remains unconfirmed.
- Import original Isovar 1.8.1 records for PacBio T1, ONT T1/T2/T3, short-read single-cell T1/T2/T3, and bulk T0/T2. Retain exact record/header hashes, qualities, flags, barcode/UMI tags, provenance and multiplicity. Inventory all 164 upstream RNA-product outcomes and the remaining vaccine variants.
- Use **complete original regional BAMs**, not the selected regression subsets, for the actual model report. This matters: the full ONT T1/T2 default contexts lack the entire long-mRNA window although the selected fixtures contain it. No thresholds are changed to force agreement. The earliest bulk T0 context supports all four documented native windows and anchors the final constructs.
- Cache real NetMHCpan 4.2c and Pepsickle 0.1.3 outputs for nine native contexts plus four documented constructs. Preserve all requested five non-null class-I alleles, lengths 8–11, affinity/EL predictions, exact internal and boundary cuts, model hashes and missing-context padding. CPN is only conditional initial-terminal recognition, not serum kinetics.
- Match all emitted peptides against all annotated protein sequences in full human Ensembl 114 with source transcripts/proteins/biotypes and the versioned oncoref CTA candidate universe. A real report flag is `DTGLKQAL` (B*08:01 EL rank 0.47) in JLF/long-mRNA flanks, matching eight protein_coding and 22 NMD-labelled DYNC1H1 transcripts. This is not evidence of expression, tissue presentation, TCR cross-reactivity or toxicity.
- Add a reproducible offline comparison report linked to the complete per-bond/MHC inventory. Keep class II, remaining constructs, complete historical mRNA products, unspecified chemistry and other processing/delivery models explicitly unassessed.
[...]

PR #454, opened 2026-09-15
Allow varcode 8 so the stack installs from source

iskandr:
varcode 8.0.0 is released on PyPI, and `requirements.txt` capped `varcode>=7.0.0,<8.0.0`, so vaxrank and current varcode can't be installed together from source. Installing every OpenVax repo from its default branch has no solution:

```
Because only the following versions of varcode are available:
    varcode<7.0.0
    varcode>=8.0.0
and vaxrank==3.17.1 depends on varcode>=7.0.0,<8.0.0, we can conclude
that vaxrank==3.17.1 cannot be used.
```

Changes `varcode>=7.0.0,<9.0.0` and bumps to 3.17.2.

## Why varcode 8 is safe for vaxrank

varcode 8.0.0 changed the annotator contract, not the classes or the default biological model. Checked two ways:

- **Imports.** Every `varcode` name vaxrank imports resolves under 8.0.0: `Variant`, `VariantCollection`, `load_vcf_fast`, `MultiOutcomeEffect`, `varcode.effects.top_priority_effect`, `varcode.effects.effect_ordering.effect_priority`, `varcode.mutant_transcript.apply_variant_to_transcript`, `varcode.errors.ReferenceMismatchError`, `varcode.cli.variant_collection_from_args`.
- **Test suite.** `./test.sh` on this branch against varcode 8.0.0: **1557 passed, 8 failed** in 2m56s.

The 8 failures are not varcode-related. The same 8 fail with varcode 7.3.0, they come from mhctools requiring `MeasurementContext` for allele-free prediction kinds, and they are filed as #453. Everything else, including the B16 smoke test, passes on varcode 8.

`./lint.sh` passes.

## Merge order

**#449 also sets `__version__ = "3.17.2"`.** Because both branches write the same string, git merges them *without a conflict*, and two different releases ship under one version number. Whichever merges second needs a new number; this is the one case here that won't announce itself.

#455 and #456 also bump (3.17.3, 3.17.4) and conflict loudly, which is safe. Suggested order, which moves only one PR: #454 (3.17.2) → #455 (3.17.3) → #456 (3.17.4) → #449 re-pointed to 3.17.5 → #437 last, after a rebase.

Two corrections to the above, from review: the widened pin resolves to varcode **8.0.1**, not 8.0.0 (same result, 1557 passed / 8 pre-existing failures), and varcode 7.x is still on PyPI, so `pip install vaxrank` was never broken by this — the cap only blocked source installs of the whole stack.

Related: openvax/sercol#5 fixes the other source-install blocker. With both, all twelve repos resolve together.


iskandr:
**Verdict:** merge after fixes. The varcode 8 compatibility argument holds — I checked it beyond the suite and it survives — but the version number collides silently with #449 and the pin's comment now describes a version the pin no longer restricts you to.

## Verified

Ran with this branch on `PYTHONPATH` against varcode checkouts at the exact release tags.

- Suite on this branch, varcode **8.0.0**: `8 failed, 1557 passed` in 2m29s. Exactly the numbers in the body.
- Suite on this branch, varcode **7.3.0**: `8 failed, 1557 passed`. Identical failure set.
- The 8 are all `tests/test_epitope_dsl.py::test_every_allele_free_kind_reaches_every_patient_allele[...]`, each `ValueError: <kind> predictions require MeasurementContext`. Same eight under both varcode versions, so this PR causes none of them — that is #453, and it is the only thing red in CI here.
- `ruff check vaxrank tests release_upload.py` — clean.
- B16 smoke on varcode 8.0.0 with every output format (`--output-ascii/html/pdf/xlsx/neoepitope/json/csv/passing-variants`): exit 0, all files written.

## Is a green suite enough? No — here is the gap, and why it closes anyway

The one place varcode 8 actually moved code underneath vaxrank is `Variant.effect_on_transcript`. In 7.3.0 it had a hard structural-variant branch and otherwise called `_predict_variant_effect_on_transcript_raw` directly, bypassing the annotator registry. In 8.0.0 both are gone: it calls `predict_variant_effect_on_transcript` → `resolve_annotator(None)`. vaxrank calls it at `vaxrank/mutant_protein_fragment.py:699` and `vaxrank/external_input.py:103`, and `tests/test_varcode_effects.py` only covers that with a fake `_Variant` returning a canned object — so the suite would not have caught a change there.

So I ran a differential instead: every varcode entry point vaxrank touches (`load_vcf_fast`, `Variant.effects`, `EffectCollection.top_priority_effect`, `Variant.effect_on_transcript`, `effect_priority`, `apply_variant_to_transcript`) plus vaxrank's own `select_varcode_effect_outcome` / `summarize_varcode_effect_outcomes` / `MutantProteinFragment.predicted_effect`, under 7.3.0 and 8.0.0, compared field by field:

- B16 VCF — 5 variants, 18 `effect_on_transcript` calls: **identical on every field**.
- The paths B16 never reaches: CFTR splice donor `7:117531115 G>A`, CFTR splice acceptor `7:117530898 G>A`, BRCA1 exonic splice site `17:43082570 CCT>GGG` — **identical**. Same `SpliceOutcomeSet` wrapping, same highest-priority and most-likely collapses, same `predicted_effect` in both `highest_priority` and `multi_outcome` modes (e.g. `ExonSkipping:exon-skip ENSE00000718611 (p.E92_K163del)` from both).
- A structural variant (`DEL 7:117530000-117560000`) — **identical**: `LargeDeletion:sv-deletion` from `effects()` and from `effect_on_transcript()` on every overlapping transcript. That is the code path that literally moved, and it lands in the same place because `fast` now owns SV routing itself and the new `enumerate_splice_outcomes` wrapper is a pass-through for anything that is not a `SpliceSite`.

The three changelog items worth worrying about are all unreachable from vaxrank:

- **Annotator selection** — vaxrank never passes `annotator=` and never calls `use_annotator`. Neither do isovar, topiary or mhctools (grepped all four packages).
- **Germline context** — `annotate_with_context` is only reached when a caller passes a non-empty `germline=`. The string `germline` does not appear in `vaxrank/` at all.
- **`NotImplemented` → `Unresolved`** — only partial/experimental annotators return the sentinel. `FastEffectAnnotator.annotate_on_transcript` never does, so `Unresolved(mechanism="unsupported_annotation")` cannot reach vaxrank's effect handling on the default path. `MultiOutcomeEffect` is still exported from `varcode/__init__.py`, so the `isinstance` check at `vaxrank/varcode_effects.py:18` still discriminates.

Residual risk is a latent coupling, not a live bug: under 7.3.0 those two `effect_on_transcript` call sites were immune to the process-global registry default; under 8.0.0 they follow it. Nothing in the stack sets it today.

## Problems

1. **`vaxrank/version.py:1` — 3.17.2 collides with #449, and the collision merges without a conflict.** #449 (`fix/440-sdist-test-policy`) is also `__version__ = "3.17.2"`. Because both branches write the *identical* line, git merges them silently:

   ```
   $ git merge-tree --write-tree origin/fix/440-sdist-test-policy origin/cli/allow-varcode-8
   bbf398a443211bc1738d9d05549c8a5a4016be57          # no CONFLICT reported
   $ git show bbf398a:vaxrank/version.py
   __version__ = "3.17.2"
   $ git ls-tree bbf398a tests/test_sdist_contents.py
   100644 blob 80db771...  tests/test_sdist_contents.py   # #449's change is in there too
   ```

   That tree carries both PRs' changes under one version number, and nothing warns you — so two different code states would ship as 3.17.2. #455 (3.17.3) and #456 (3.17.4) each *do* conflict in `version.py` against every sibling, which is the loud, safe behavior. Either this PR or #449 has to move.

2. **`requirements.txt:4-5` — the comment above the pin is now stale.** It still reads "Varcode 7 supplies splice outcome sets unconditionally and defaults to fast annotation. Vaxrank uses its keyword-free effect API and outcome selection." The pin now admits 8.x, where `fast` is the *only* built-in annotator and owns structural routing too, and where the plugin contract is `NotImplemented`/`Unresolved`. Every other pin in this file carries a comment justifying its floor; this one describes a constraint the line no longer expresses. Suggest something closer to: floor 7.0.0 for the always-on splice outcome sets and the `fast` default; 8.x keeps both and changes only the annotator plugin contract, which vaxrank does not use.

## Nits

- **The pin resolves to 8.0.1, not 8.0.0.** `varcode>=7.0.0,<9.0.0` resolves to `varcode==8.0.1` (`uv pip compile`), which is the current PyPI release — so this pin change moves every fresh vaxrank install from 7.3.0 straight to 8.0.1, and the body only claims 8.0.0. I ran it: suite against **8.0.1** is `8 failed, 1557 passed`, the same #453 set, and every name vaxrank imports resolves. 8.0.1 only renames the experimental `realized` annotator to `transcript_model`, is purely additive in `varcode/__init__.py`, and does not touch `variant.py`, `effects/`, `mutant_transcript.py`, `errors.py` or `cli/`. Worth naming 8.0.1 in the body so the changelog records what users will actually get.
- The body's opening — "vaxrank and current varcode can't be installed together from source" — reads like a PyPI breakage on a skim. It isn't: varcode 7.0.0–7.3.0 are all live and unyanked, and `vaxrank==3.17.1` resolves today to `varcode==7.3.0`. The quoted resolver output is specific to the all-repos-from-default-branch case, which the next sentence does say. A small rewording would keep the changelog from recording a break that never happened.
- Nothing pins or tests the new `<9.0.0` ceiling, so a varcode 9 with another contract change would install silently. Consistent with the rest of the file — noting only that the cap is documentation, not a guard.

## Merge order

`vaxrank/version.py` is the only file all of #437, #449, #454, #455 and #456 touch. Suggested order, which moves only one PR's number:

1. **#454 (this one), 3.17.2** — first, since it unblocks a source install of the whole stack.
2. **#455, 3.17.3** — unchanged.
3. **#456, 3.17.4** — unchanged.
4. **#449 → 3.17.5** — must move off 3.17.2, and it is the one whose collision is silent, so it is the safest to re-point.
5. **#437, 3.18.0** — last. It is already `CONFLICTING` against `main`, and it add/add-conflicts with all three of these on `tests/osteosarc_selection_helpers.py` and `tests/data/osteosarc/selection_validation/import_assets.py`; its `requirements.txt` hunk (`isovar>=1.8.0` → `1.8.1`) is already a no-op on current `main`.

Beyond `version.py`, the three `cli/*` branches do not overlap: this PR touches only `requirements.txt`, #455 touches `report.py:188-196` and `entry_point.py:1633`, #456 touches `report.py` imports plus the writers at 850/858/893/1020/1095/1100. The only non-version file-level overlap in the set is #437's `report.py` hunk at `_pdf_via_pdfkit` (line ~872) sitting next to #456's `make_pdf_report` change at 893 — another reason to land #437 last.

iskandr:
Pushed `b756512`, which addresses **problem 2** (stale pin comment): it now states why 7.0.0 is the floor (always-on splice outcome sets, `fast` as the default annotator) and why 8.x is admitted — the annotator-contract change is the `NotImplemented` / `Unresolved` plugin protocol and germline context, none of which vaxrank reaches, since it selects no annotator and passes no germline context.

Re-ran after the change: `ruff check vaxrank tests release_upload.py` clean, `tests/test_cli_flags.py tests/test_entry_point_helpers.py` 37 passed. I also confirmed the comment lines in `requirements.txt` don't leak into the built metadata — `setup.py` feeds all 59 lines to `install_requires`, 40 of which are comments, but setuptools drops them (published vaxrank 3.17.1 has 21 clean `Requires-Dist` entries, 0 comment-derived). So no packaging change needed.

Left for you:

- **Problem 1 — the 3.17.2 collision with #449.** This is a merge-order decision, and it is the one that needs a human: whichever of the two you want to land first keeps 3.17.2 and the other moves. Not fixed here because re-pointing this branch would just push the silent collision onto #449. My suggested order is in the review — this PR first at 3.17.2, #455 at 3.17.3, #456 at 3.17.4, #449 re-pointed to 3.17.5, #437 last at 3.18.0 after a rebase.
- Both nits (naming 8.0.1 in the body, and the opening sentence reading like a PyPI breakage) are PR-description wording, so I left the body alone. The verification for 8.0.1 is in the review comment: suite is `8 failed, 1557 passed`, same #453 set.

iskandr:

```
```

iskandr:
The current Varcode 9.3.7 compatibility failure is tracked in #499 and fixed in PR #498 with varcode>=7.0.0,<10.0.0. The original <9.0.0 proposal here would still exclude 9.3.7. The Varcode-9 API audit found Vaxrank does not use the removed annotator/plugin interfaces; full-suite and B16 validation against published 9.3.7 is underway. Keeping this PR open until the replacement lands.

iskandr:
Superseded by #498, now merged and released as Vaxrank 3.20.0. The released requirement is varcode>=7.0.0,<10.0.0, including both Varcode 8 and 9. Lint, all 1,758 local tests against published Varcode 9.3.7, B16 smoke, the Python 3.10–3.13 CI matrix, and the gated deployment passed; PyPI wheel/source files are verified.

PR #455, opened 2026-09-15
Keep internal parser state out of reports and the console

iskandr:
`parse_vaxrank_args` stashes a ~110-key snapshot of every parser default on the namespace as `_parser_defaults`. The pipeline path sets `data['args'] = vars(args)` and the template reports render that namespace verbatim, so the internal snapshot ends up in two user-facing places.

Reproduced with the B16 example (`run-vaxrank-b16-test-data.sh`):

```
$ pdftotext -layout vaccine-peptides-report.pdf - | sed -n '28,32p'
COMMAND LINE ARGUMENTS
                        {'version': '==SUPPRESS==', 'print_default_config': '==SUPPRESS==', 'vcf': [], 'maf': [], 'variant': [],
                        'genome': None, 'download_reference_genome_data': False, 'json_variants': [], 'bam': None,
                        'min_mapping_quality': 1, 'use_duplicate_reads': False, 'drop_secondary_alignments': False,
...
 _parser_defaults       'vaccine_type': ['peptide', 'mrna'], 'output_dir': '', 'force_overwrite': False, ...
```

The `_parser_defaults` row fills the whole first page of the args table in the HTML and PDF reports. The same dict also went to the console at INFO (`About to save args: {...}`), immediately after `log_args_summary` had already printed a grouped, readable summary of the resolved arguments.

Changes:

- `report.py`: drop underscore-prefixed keys from the report args table (alongside the existing `output*` / SUPPRESS / unset filters).
- `entry_point.py`: `About to save args` → `logger.debug`, so the raw namespace still lands in `--log-path` for provenance but not in the terminal.
- Regression test in `tests/test_cli_flags.py`: build the args with the real `parse_vaxrank_args`, render the packaged HTML and ASCII templates, assert no underscore key reaches `template_data['args']` and `_parser_defaults` appears in neither rendering.

Verified: `ruff check vaxrank tests release_upload.py` clean; the new test fails on `main` and passes here; re-ran the B16 pipeline with `--output-pdf-report` / `--output-html-report` and neither output nor the console mentions `_parser_defaults`.

Version bumped to 3.17.3 (3.17.2 is taken by #449, 3.18.0 by #437).


iskandr:
CI note: the `build (3.x)` failures on this branch are #453, not this change — the same 8 `tests/test_epitope_dsl.py::test_every_allele_free_kind_reaches_every_patient_allele[...]` cases (`ValueError: <kind> predictions require MeasurementContext`) reproduce on an unmodified `origin/main` checkout with the current mhctools 3.44.10 / topiary 5.57.0:

```
$ git -C vaxrank status --short   # clean, at origin/main 84f9ebc
$ pytest tests/test_epitope_dsl.py -k allele_free_kind -q
8 failed, 7 passed, 83 deselected
```

Nothing in this diff touches `epitope_dsl.py` or its tests.

iskandr:
**Verdict:** merge after fixes. The leak is real, the fix is in the right layer, and every claim in the body checks out — but the filter is one character broader than it needs to be and takes real provenance out of LENS reports with it.

## Verified

- The new `tests/test_cli_flags.py::test_report_args_table_omits_internal_parser_state` passes here and fails against `main` — ran this branch's test file against the `main` source tree: `assert all(not key.startswith("_") for key, _ in template_data["args"])` → `assert False`.
- `ruff check vaxrank tests release_upload.py` — clean.
- Re-ran the B16 pipeline on this branch with `--output-ascii-report --output-html-report --output-pdf-report --output-json-file`: exit 0, and `_parser_defaults` occurs **0** times in the console output, **0** in the ASCII report, **0** in the HTML report, and is absent from the PDF bytes. Against `main` the same `TemplateDataCreator` construction puts it in the table (71 rows there vs 69 here).
- The `logger.debug` claim holds. `vaxrank/logging.conf` sets `logger_vaxrank` to `DEBUG` and `handler_fileHandler` to `DEBUG` while `handler_consoleHandler` is `INFO`, so `Saving args` still lands in the run log — 1 occurrence in `python.log` from the run above — and never on the terminal.
- Fixing it in `report.py` rather than at the call site is the right call: the `--input-json-file` replay path (`entry_point.py:1563` `data['args'].update(vars(args))`, then `args_for_report = data['args']` at 1245) goes through the same filter, so re-rendering an old JSON is cleaned too.

## Problems

1. **`vaxrank/report.py:191` drops every underscore key, and that takes `_inferred_mhc_alleles_from_lens` with it.**

   There are exactly two underscore keys on the namespace today: `_parser_defaults` and `_inferred_mhc_alleles_from_lens`, the latter set at `entry_point.py:1220` on the LENS / pVACseq path — where `args_for_report = vars(args)` (line 1239), so it does reach the report table.

   `log_args_summary` already distinguishes the two on purpose. `entry_point.py:519`:

   ```python
   inferred_rows = [
       (k, v) for k, v in sorted(vars(args).items())
       if k.startswith('_') and k != '_parser_defaults' and v
   ]
   ```

   i.e. the console has a dedicated auto-wired block precisely so an operator can see which alleles vaxrank inferred rather than was given. This PR removes that same fact from the report:

   ```
   # built args via parse_vaxrank_args([...]), then
   #   args._inferred_mhc_alleles_from_lens = ["HLA-A*02:01", "HLA-B*07:02"]
   # and ran TemplateDataCreator(...).compute_template_data()

   main          n_args_rows 71  _parser_defaults: True   _inferred_mhc_alleles_from_lens: True
   this branch   n_args_rows 69  _parser_defaults: False  _inferred_mhc_alleles_from_lens: False
   ```

   On a LENS run the report's args table has no `--mhc-alleles` row (it was never passed) and now no record of what was auto-wired either — which is exactly what a reviewer of that report would want to check. Suggest mirroring the rule the console already uses: exclude `_parser_defaults` specifically, or keep non-empty underscore keys. The test at `tests/test_cli_flags.py:326` asserts the broad form, so it would need to move with it — otherwise it locks this in.

2. **`--output-json-file` still carries the full snapshot.** `entry_point.py:1639` serializes `data['args']` verbatim, so `ranked_vaccine_peptides.json` still contains `_parser_defaults` with all 124 keys — 3,697 of the 46,595 bytes in the B16 JSON from the run above:

   ```
   keys in args: 124
   _parser_defaults in JSON args: True  (nested keys: 124)
   json size: 46595 bytes;  serialized _parser_defaults: 3697 bytes
   ```

   The JSON is a documented primary output (`--output-json-file`, "machine-readable JSON of ranked vaccine peptides"), and what it records there is the *defaults table*, not what ran — so as provenance it is actively misleading, which is the same argument the PR makes for the args table. Either strip it at the dump too, or state in the body that the JSON is deliberately the provenance sink (which is what the new `logger.debug` comment implies the log file is for — the two can't both be the answer).

## Nits

- `logger.debug('Saving args: %s', data['args'])` uses lazy `%s` formatting, so there's no cost when DEBUG is off. Fine as is.
- The guard at `tests/test_cli_flags.py:301` — `assert "_parser_defaults" in vars(args), "test no longer exercises the leak"` — is a good touch; keep it whichever way problem 1 lands.
- The comment block at `report.py:173-183` is now doing a lot of work for one filter expression. Not asking for a change; just noting that if problem 1 narrows the rule, the comment should name `_parser_defaults` rather than "underscore-prefixed internal state", since `_inferred_mhc_alleles_from_lens` is not internal state.

## Merge order

`vaxrank/version.py` is the only file all of #437, #449, #454, #455 and #456 touch, and this PR's 3.17.3 conflicts with each sibling's version line — which is the safe, visible failure mode. (The dangerous pair is #454 and #449: both write `__version__ = "3.17.2"`, so they merge into each other with **no** conflict and would ship two code states as one release.)

Suggested order, which re-points only one PR:

1. **#454, 3.17.2** — first; it unblocks a source install of the whole stack.
2. **#455 (this one), 3.17.3** — unchanged.
3. **#456, 3.17.4** — unchanged.
4. **#449 → 3.17.5** — must move off 3.17.2.
5. **#437, 3.18.0** — last; already `CONFLICTING` against `main` and add/add-conflicts with all three `cli/*` branches on `tests/osteosarc_selection_helpers.py` and `tests/data/osteosarc/selection_validation/import_assets.py`.

Content-wise this PR is independent of the other two: #454 touches only `requirements.txt`, and #456's `report.py` edits are the import block plus the writers at 850/858/893/1020/1095/1100 — nowhere near this PR's `report.py:188-196`. Whichever of #455/#456 goes second needs only the version line nudged.

iskandr:
Pushed `22af986`, which addresses **problem 1**.

`_parser_defaults` moves into `_internal_keys` and the blanket `not k.startswith("_")` clause is replaced with "keep underscore keys that hold a value" — the same rule `log_args_summary` applies to its auto-wired console block (`entry_point.py:519`). So the ~110-key snapshot still never reaches a report, and `_inferred_mhc_alleles_from_lens` is back in the args table on LENS / pVACseq runs. An empty one means nothing was inferred and stays out.

`test_report_args_table_omits_internal_parser_state` no longer asserts the blanket "no underscore key" shape it was locking in; it asserts what the PR is actually about (`_parser_defaults` absent from the table and from both renderings). A second test, `test_report_args_table_keeps_auto_wired_provenance`, pins the other half in both directions.

Verified after the change:

- `ruff check vaxrank tests release_upload.py` — clean.
- Full suite: `8 failed, 1559 passed` — two more passing than before (the split test), same eight `test_epitope_dsl.py` `MeasurementContext` failures that are #453 and fail identically on `main` and on `varcode` 7.3.0 / 8.0.0.
- Behavior check through the real `parse_vaxrank_args` → `TemplateDataCreator`: `_parser_defaults` absent, `_inferred_mhc_alleles_from_lens` present, 70 rows (vs 71 on `main` with the leak, 69 with the over-broad filter).

Left for you:

- **Problem 2 — `--output-json-file` still carries `_parser_defaults`** (3,697 of 46,595 bytes on B16). That's a behavior choice about what the JSON is for: either it is the provenance sink the new `logger.debug` comment implies the log file is, or it is a user-facing report and should be filtered like the others. Either is defensible, so I left it rather than pick for you — but the PR body should say which, since right now the JSON and the log file are both described as the place the raw namespace survives.
- The version bump is untouched at 3.17.3. Note it conflicts in `version.py` with #454, #456 and #449 — which is the safe, visible failure. The pair to watch is #454 + #449: both write `__version__ = "3.17.2"`, so they merge with no conflict at all.

iskandr:

```
```

PR #456, opened 2026-09-15
Create missing output directories before writing pipeline outputs

iskandr:
`--output-dir DIR` is the documented way to get the full bundle, and `populate_default_output_paths` fills in `DIR/ranked_vaccine_peptides.{csv,json}` and `DIR/vaccine_report.{txt,pdf}`. Only the construct writers (`_emit_peptide_constructs` / `_emit_mrna_constructs` / `write_run_summary`) called `os.makedirs`, and the JSON dump runs before them, so the flow crashes when `DIR` does not exist yet — after the whole pipeline has already run:

```
$ vaxrank --vcf tests/data/b16.f10/b16.vcf --bam tests/data/b16.f10/b16.combined.bam \
    --mhc-predictor random --mhc-alleles H2-Kb,H2-Db --mhc-peptide-lengths 8 \
    --vaccine-type peptide --output-dir out-pep
...
  File ".../vaxrank/cli/entry_point.py", line 1634, in ranked_vaccine_peptides_with_metadata_from_parsed_args
    with open(args.output_json_file, 'w') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'out-pep/ranked_vaccine_peptides.json'
```

`mkdir out-pep` first and the same command succeeds. The same class of failure hits an explicit path in a directory that does not exist yet:

```
$ vaxrank ... --output-csv subdir/x.csv
OSError: Cannot save file into a non-existent directory: 'subdir'
```

The external-input path already handled this deliberately (`epitope_io._ensure_parent_dir`, "so `--output-csv foo/bar.csv` works without the user pre-creating `foo/`"); the pipeline writers just never used it.

Changes:

- `epitope_io._ensure_parent_dir` → public `ensure_parent_dir` (docstring extended to cover `open(path, 'w')` and the auto-populated `--output-dir` paths).
- Call it from the writers that lacked it: JSON dump, passing-variants CSV, Isovar CSV, `save_predictions`, ranked CSV / XLSX, neoepitope XLSX, and the ASCII / HTML / PDF reports.
- Two tests in `tests/test_entry_point_helpers.py`: the helper's nested / bare-name / idempotent behavior, and a template report written into a directory that does not exist yet.

Verified: `ruff check vaxrank tests release_upload.py` clean; `pytest tests/test_entry_point_helpers.py tests/test_epitope_io.py tests/test_neoepitope_report.py tests/test_external_input.py tests/test_cli_flags.py` → 238 passed; re-ran the B16 pipeline with `--output-dir brand/new/dir --output-passing-variants-csv nested/passing.csv --output-epitopes nested2/ep.tsv` and all three trees are created with the expected files (exit 0).

Version bumped to 3.17.4 (3.17.2 is taken by #449, 3.17.3 by #455, 3.18.0 by #437).


iskandr:
CI note: the `build (3.x)` failures on this branch are #453, not this change — the same 8 `tests/test_epitope_dsl.py::test_every_allele_free_kind_reaches_every_patient_allele[...]` cases (`ValueError: <kind> predictions require MeasurementContext`) reproduce on an unmodified `origin/main` checkout with the current mhctools 3.44.10 / topiary 5.57.0:

```
$ git -C vaxrank status --short   # clean, at origin/main 84f9ebc
$ pytest tests/test_epitope_dsl.py -k allele_free_kind -q
8 failed, 7 passed, 83 deselected
```

Nothing in this diff touches `epitope_dsl.py` or its tests.

iskandr:
**Verdict:** merge after fixes. The coverage is complete — every pipeline output path really is handled, including the `--output-dir` bundle — but the two new tests miss the exact crash the PR is named after, and the fix stays lazy where the repo already has a pre-flight hook for this.

## Verified

Reproduced both failures on `main` with the B16 data, then re-ran on this branch.

**On `main`:**

```
--output-csv missing-dir/out.csv
  OSError: Cannot save file into a non-existent directory: 'missing-dir'
  raised from report.py:1090  all_dfs.to_csv(...)   — after 16.7 s of pipeline

--output-dir main-out/brand/new
  FileNotFoundError: [Errno 2] No such file or directory:
      'main-out/brand/new/ranked_vaccine_peptides.json'
```

Both exactly as described in the body.

**On this branch:** one run with all nine explicit output flags aimed at nine different non-existent directories — `missing-dir/out.csv`, `iso/nested/iso.csv`, `ep/dir/ep.tsv`, `pv/pv.csv`, `xl/rep.xlsx`, `neo/neo.xlsx`, `js/out.json`, `h/rep.html`, `a/rep.txt` — exit 0, all nine trees created, all nine files written. A second run with `--output-dir br-out/brand/new --output-pdf-report pdfdir/rep.pdf` → exit 0 and the full bundle: `ranked_vaccine_peptides.{csv,json}`, `vaccine_report.txt`, `run_summary.txt`, `peptide/{vaccine.fasta,manifest.json,order_form.csv,vaccine_report.{txt,pdf}}`, `mrna/{full,cds,no_polyA}.fasta` + manifest + parts CSV, plus the nested PDF.

**Coverage audit — yes, every output path, not just some.** ASCII/HTML/PDF at `report.py:850/858/893`; XLSX at `1020` and `1100`; ranked CSV at `1095`; JSON at `entry_point.py:1637`; isovar CSV at `1513`; passing-variants CSV at `1596`; `save_predictions` at `epitope_io.py:261`; neoepitope CSV/XLSX at `epitope_io.py:1557/1562` (already had it); and the `--output-dir` modality subdirs via the pre-existing `os.makedirs` at `entry_point.py:377/749/998/1409`. I grepped every `to_csv` / `to_excel` / `ExcelWriter` / `open(..., 'w')` / `write_text` in the package — the only writers still without it are `construct_audit.py`, `cleavage_report.py` and `context_report.py`, none of which has a CLI caller.

Also: both new tests fail against the `main` source and pass here; `ruff check vaxrank tests release_upload.py` clean; and `report.py` importing `epitope_io` introduces no cycle (`epitope_io` does not import `report`).

## Problems

1. **Neither new test covers the crash this PR is named after, and there is no `--output-dir` test.**

   `tests/test_entry_point_helpers.py:530` covers `make_ascii_report` / `make_html_report`. `tests/test_entry_point_helpers.py:518` only exercises the helper itself — which already existed and already worked; it fails on `main` purely because of the `_ensure_parent_dir` → `ensure_parent_dir` rename, not because of any behavior this PR changed. So of the ten call sites added, seven are unguarded: the JSON dump at `entry_point.py:1637` (the traceback quoted in the body), `make_csv_report` at `report.py:1095`, the XLSX writers at `1020`/`1100`, `save_predictions`, the isovar CSV and the passing-variants CSV. Nothing would fail if any of them were reverted.

   A parametrize over the writers, each handed a `tmp_path / "missing" / name`, would cover all of them in roughly the lines the two current tests take. The `--output-dir` case specifically deserves one too, since that is the path `populate_default_output_paths` auto-fills and the one the body leads with — a test at the `emit_outputs` / `ranked_vaccine_peptides_with_metadata_from_parsed_args` level with a `SimpleNamespace` args and an empty result list would do it without running a pipeline.

2. **The fix is entirely lazy, so a bad output path still fails only at the end.** That is the motivation in the body — "after the whole pipeline has already run" — and this PR removes exactly one instance of it. The others remain: an unwritable parent, or a path whose directory component is a regular file, still blows up in the writer. That last one now raises a different exception at the same late point:

   ```python
   >>> open("somefile", "w").write("x")
   >>> ensure_parent_dir("somefile/out.csv")
   FileExistsError: [Errno 17] File exists: '.../somefile'
   ```

   `arg_parser.check_args` (`arg_parser.py:809`) already exists as the pre-flight hook for precisely this class of mistake — it is where `_reject_output_dir_misuse` catches `--output-dir out.fasta` so the user "sees the mismatch immediately rather than burning ranking time and crashing in the writer", in its own words. Creating (or at least probing) the output parents there would fix the reported crash *and* fail in the first second. Your call whether that belongs here or in a follow-up, but as written the PR moves one failure and leaves the rest of the class in place.

3. **Yes, a mistyped path now silently makes a junk tree — on every output flag.** `--output-csv /tmp/typo/out.csv` creates `/tmp/typo`; `--output-dir a/b/c/d` creates the whole chain. In the branch run above, `missing-dir/`, `iso/nested/`, `ep/dir/`, `pv/`, `xl/`, `neo/`, `js/`, `h/`, `a/` and `pdfdir/` were all materialized from single command lines.

   This is the same trade `epitope_io._ensure_parent_dir` already took deliberately for the external path, so the PR is consistent rather than novel — but it is now the behavior for the *primary* outputs including `--output-dir`, and `confirm_output_dir_overwrite` (`entry_point.py:1005`) only warns about a directory that already exists and is non-empty, so it never sees a typo'd one. Cheapest mitigation: a `logger.info("Created output directory %s", parent)` inside `ensure_parent_dir` when it actually creates something, so the created path is at least visible in the run log and the summary, plus a line in the `--output-dir` / `--output-csv` help text saying missing directories are created.

## Nits

- `--log-path` is the same class and is still broken: `--log-path nolog/vaxrank.log` dies with a raw traceback out of `logging.config.fileConfig` (`entry_point.py:1045`) — `FileNotFoundError: [Errno 2] No such file or directory: '.../nolog/vaxrank.log'`. It fails fast rather than after the pipeline, so it costs far less than what this PR fixes, but the user gets a stack trace instead of a message.
- `ensure_parent_dir` is now a generic path helper imported by `report.py` from `epitope_io`, which is a heavy module (mhctools, external prediction, the DSL). No cycle and `report.py` already imports it, so nothing is broken — but a two-line `vaxrank/paths.py` would be a more natural home now that it is public and used from three modules.
- `vaxrank/epitope_io.py:1412` — the widened docstring is good. Worth one more clause saying it creates *arbitrarily deep* parents, since that is the half that surprises people (and is problem 3).
- `make_pdf_report` creates the directory at `report.py:893` before rendering, so a render failure leaves an empty directory behind. Harmless, just noting the ordering.

## Merge order

`vaxrank/version.py` is the only file all of #437, #449, #454, #455 and #456 touch. This PR's 3.17.4 conflicts with each sibling's version line, which is the safe, visible failure. (The dangerous pair is #454 and #449: both write `__version__ = "3.17.2"`, so they merge with **no** conflict and would ship two different code states as one release.)

Suggested order, which re-points only one PR:

1. **#454, 3.17.2** — first; it unblocks a source install of the whole stack.
2. **#455, 3.17.3** — unchanged.
3. **#456 (this one), 3.17.4** — unchanged.
4. **#449 → 3.17.5** — must move off 3.17.2.
5. **#437, 3.18.0** — last, after a rebase. It is already `CONFLICTING` against `main`, add/add-conflicts with all three `cli/*` branches on `tests/osteosarc_selection_helpers.py` and `tests/data/osteosarc/selection_validation/import_assets.py`, and — relevant here — its only `report.py` hunk lands at `_pdf_via_pdfkit` (line ~872), immediately above this PR's `make_pdf_report` change at 893. That is the one real content overlap in the whole set, so #437 going last keeps the conflict on its side.

Against #455 this PR is content-independent: #455 edits `report.py:188-196` (the args-table filter) and `entry_point.py:1633`, neither of which this touches. Whichever of the two goes second needs only its version line nudged.

iskandr:
Pushed `f04df2e`, which addresses **problem 1** and the visibility half of **problem 3**.

**Problem 1 — tests.** One case per writer, so removing any single `ensure_parent_dir` call now fails a test:

- `test_tabular_writers_create_their_parent_directory` — parametrized over the ranked CSV, the ranked XLSX, the minimal neoepitope XLSX and `save_predictions`.
- `test_isovar_csv_creates_its_parent_directory` — drives the real `run_vaxrank_from_parsed_args` with the stub set `tests/test_isovar_config.py` already uses.
- `test_json_dump_creates_its_parent_directory` — drives `ranked_vaccine_peptides_with_metadata_from_parsed_args`, covering the traceback in the description plus the passing-variants CSV.
- `test_output_dir_bundle_is_written_into_a_missing_directory` — the `--output-dir` case end to end: `populate_default_output_paths` fills the paths inside a `brand/new/dir` that does not exist, then the JSON dump writes into it.

All nine tests in that section fail against the `main` source, and the new ones fail with the real `OSError: Cannot save file into a non-existent directory` rather than an import error — so `test_ensure_parent_dir_...` is no longer the only "fails on main because of the rename" case.

**Problem 3 — junk trees.** Not preventable without the pre-flight change (problem 2), but no longer silent: `ensure_parent_dir` logs `Created output directory <path>` when it actually creates one, so a typo shows up in the run log instead of quietly materializing a tree. Confirmed on B16 with `--output-dir fresh/tree --output-csv typo-dir/out.csv`:

```
INFO - vaxrank.epitope_io:1431 - Created output directory fresh/tree
INFO - vaxrank.epitope_io:1431 - Created output directory typo-dir
```

The docstring now says outright that intermediate components are created too, which is the surprising half.

Verified after the change: `ruff check vaxrank tests release_upload.py` clean; full suite `8 failed, 1566 passed` (seven more passing than the 1559 baseline — the new tests; the eight failures are the `test_epitope_dsl.py` `MeasurementContext` set from #453, unchanged); `tests/test_entry_point_helpers.py tests/test_epitope_io.py tests/test_neoepitope_report.py tests/test_external_input.py tests/test_cli_flags.py` 245 passed; B16 end-to-end with `--output-dir fresh/tree --output-csv typo-dir/out.csv` exits 0 with the full bundle.

Left for you:

- **Problem 2 — pre-flight vs lazy creation.** Whether to create/probe the output parents in `arg_parser.check_args` so a bad path fails in the first second instead of after the run is a design call about where that belongs, and it would change what `check_args` is responsible for. Left alone. As it stands an unwritable parent, or a path whose directory component is an existing file (`FileExistsError` out of `os.makedirs` now, rather than pandas' `OSError`), still fails at the end.
- **`--log-path` into a missing directory** still dies with a raw traceback from `logging.config.fileConfig`. It is outside "pipeline outputs" and it fails fast, so I left it — worth a separate issue if you want the message cleaned up.
- The version bump is untouched at 3.17.4.

iskandr:

```
```

PR #477, opened 2026-09-16
Require Isovar 1.17 read identity and preserve RNA evidence contracts

### LINKED ISSUE #476: Update RNA-context contracts for Isovar #264 read-identity correction
Isovar #264 corrects sequenced-read identity, mate merging, support counts and assembly across alternative SAM placements. During downstream verification, three Vaxrank RNA-context assertions were found to rely on an ambiguous H1-2 bulk template.

In the pinned original T0 RNA, read2 of `...2382:6460:24392` has primary `99M15D46M4S` and secondary `118M31S` placements with conflicting allele calls. Isovar 1.17.0 retains these as uncertain rather than definitive deletion support. One unambiguous merged pair remains (two segments, one template), below the unchanged coverage-two floor; there is no default protein. Coverage-one remains an explicit diagnostic opt-in, not a default change.

The Vaxrank runtime already handles missing Isovar proteins. Update its minimum Isovar version and regressions, preserving API shapes, ranking policy, ambiguity, and the absolute coverage floor. Ship after Isovar 1.17.0 is available on PyPI.

iskandr:
## Summary

Fixes #476; companion to openvax/isovar#283 / openvax/isovar#264. Version 3.18.1.

- Require Isovar >=1.17.0 so sequenced segments, alternative alignments and real mates are not conflated.
- Correct the old bulk-T0 H1-2 regression contract: a second template has competing deletion/nondeletion placements, leaving one unambiguous merged pair. No default protein at the unchanged two-observation floor; missing-protein handling already works. Explicit coverage-one/24-aa diagnostics remain opt-in. No runtime API, ranking, ambiguity/provenance, or default-policy changes.
- Regenerate the real local NetMHCpan 4.2c cache with Isovar 1.17.0 and Vaxrank 3.18.1. All 11 contexts, 614 peptides and all 6,140 complete prediction rows are identical after matching row keys. Only provenance and row ordering changed; this is an actual model rerun, not relabeling old evidence. The cache's H1-2 ONT-T2 fixture is distinct from the uncertain bulk-T0 fixture.

## Verification

- `./lint.sh`: passed.
- `./test.sh`: 1,605 passed against both the Isovar 1.17.0 branch and its actual published PyPI wheel (isolated target; import path verified).
- Real NetMHCpan rerun and exact dataframe comparison passed. TSV terminal tabs are intentional empty schema columns, not whitespace to strip.
- GitHub CI must pass with the published dependency; merge followed by `./deploy.sh` and PyPI verification.


iskandr:
Published-wheel verification is green: 1,605 tests and lint pass using Isovar 1.17.0 downloaded into an isolated target (import path verified). The initial push workflow and its first retry failed during pip dependency installation because their PyPI index response still listed only <=1.16.2; no test ran in those jobs. The separate PR workflow has resolved 1.17.0 and is running/passing tests. Rerunning the failed push matrix after index propagation; no dependency floor or test gate was weakened.

iskandr:
Shipped as Vaxrank 3.18.1: https://pypi.org/project/vaxrank/3.18.1/ (tag v3.18.1, merge 649ed8e). Both CI matrices passed on Python 3.10–3.13 after the stale PyPI index expired. Clean-main deploy.sh passed lint and all 1,605 tests against the published Isovar 1.17.0 wheel, then published and verified both release distributions. No ranking/default-policy changes; cached real NetMHCpan values remain identical.

